PR #21, opened 2026-04-22
fix: create log directory before stderr redirect in hooks

dmar842:
## Problem

The `SessionStart` and `PostToolUse` hook commands in `hooks/hooks.json` redirect stderr to `${CLAUDE_PROJECT_DIR}/.remember/logs/hook-errors.log`, but nothing guarantees `.remember/logs/` exists first.

In any Claude Code project where the plugin has not yet run, bash emits a hook error like:

```
/usr/bin/bash: line 1: /c/path/to/project/.remember/logs/hook-errors.log: No such file or directory
```

The hook scripts (`session-start-hook.sh`, `post-tool-hook.sh`) never run. I hit this today on Windows but the same race exists on Linux/macOS — any fresh project activation triggers it.

## Fix

Prepend an idempotent `mkdir -p` to both hook command strings, with `2>/dev/null` so any mkdir failure (e.g. read-only filesystem) is silent and doesn't itself fail the redirect it's trying to enable.

## Diff

```diff
-  "command": "bash \"${CLAUDE_PLUGIN_ROOT}/scripts/session-start-hook.sh\" 2>> ..."
+  "command": "mkdir -p \"${CLAUDE_PROJECT_DIR:-.}/.remember/logs\" 2>/dev/null; bash \"${CLAUDE_PLUGIN_ROOT}/scripts/session-start-hook.sh\" 2>> ..."
```

Same change applied to both `SessionStart` and `PostToolUse` commands.

## Test

Reproduced the bug in a Claude Code project dir without `.remember/`, then applied the patch manually and confirmed the hook error disappears on the next `PostToolUse` fire and `hook-errors.log` is created empty as expected.

### DIFF
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -5,7 +5,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "bash \"${CLAUDE_PLUGIN_ROOT}/scripts/session-start-hook.sh\" 2>> \"${CLAUDE_PROJECT_DIR:-.}/.remember/logs/hook-errors.log\""
+            "command": "mkdir -p \"${CLAUDE_PROJECT_DIR:-.}/.remember/logs\" 2>/dev/null; bash \"${CLAUDE_PLUGIN_ROOT}/scripts/session-start-hook.sh\" 2>> \"${CLAUDE_PROJECT_DIR:-.}/.remember/logs/hook-errors.log\""
           }
         ]
       }
@@ -15,7 +15,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "bash \"${CLAUDE_PLUGIN_ROOT}/scripts/post-tool-hook.sh\" 2>> \"${CLAUDE_PROJECT_DIR:-.}/.remember/logs/hook-errors.log\""
+            "command": "mkdir -p \"${CLAUDE_PROJECT_DIR:-.}/.remember/logs\" 2>/dev/null; bash \"${CLAUDE_PLUGIN_ROOT}/scripts/post-tool-hook.sh\" 2>> \"${CLAUDE_PROJECT_DIR:-.}/.remember/logs/hook-errors.log\""
           }
         ]
       }


0reo:
Also hit this on Linux today in a project where `.remember/logs/` didn't exist yet — not a Windows-specific issue. The failure chain is self-reinforcing: the `SessionStart` redirect fails, so `session-start-hook.sh` never runs, so the `mkdir` at lines 57–59 of that script never executes, and then every subsequent `PostToolUse` in the session fails the same redirect. No recovery without manual intervention.

Confirmed the bash behavior locally: `bash -c 'true 2>> /nonexistent/x.log'` exits `1` with `true` never executing — the shell opens redirect targets before running the command, so a failed open aborts the whole command.

See also #23, which is an independent report of the same bug filed ~9 hours after this PR went up.

The fix looks right. One small thing worth considering: `mkdir -p "${CLAUDE_PROJECT_DIR:-.}/.remember/logs" 2>/dev/null; bash ...` swallows `mkdir` errors but still attempts the `bash` — which is what you want here (the `2>>` will then fail loudly if the `mkdir` failed for some other reason like ENOSPC, and at least that signal isn't masked).

— Written by Claude Code (Opus 4.7) on behalf of @0reo

aryeh-evinced:
+1 — hitting this on remember v0.5.0 inside a Claude Code worktree (`.claude/worktrees/...`). Error seen verbatim:

```
/bin/sh: /path/to/worktree/.remember/logs/hook-errors.log: No such file or directory
```

The `mkdir -p` prepend here is the right fix — the shell resolves the `2>>` redirect target before the hook script runs, so any `mkdir` inside `scripts/log.sh` is too late.

mcmanusj-reninc:
+1 from **Windows 11 + Git Bash + Claude Code 2.1.116**. Error seen verbatim on every `PostToolUse` in a fresh project with no `.remember/` tree:

```
/usr/bin/bash: line 1: /d/projects/ren/.remember/logs/hook-errors.log: No such file or directory
```

Ran `mkdir -p /d/projects/ren/.remember/logs` locally — error disappeared immediately on the next tool call, confirming the diagnosis. Nothing Windows-specific about the bash redirect semantics here; the fix in this PR should resolve it on all three platforms (Linux per @0reo, worktree per @aryeh-evinced, Windows here).

Also filed independently as #23 by another reporter.

josemoreno801-netizen:
Heads-up — PR #28 (Windows bootstrap fix + TMPDIR portability + README prereqs) lands the same `mkdir -p .remember/logs` fix as commit 1 of its 3-commit series. Both PRs are correct and address the identical race; merging them as-is would be a no-op double-apply.

Maintainer's call on which to take:
- **Take #28** if you want the broader Windows + docs improvements alongside the hook log dir fix.
- **Take #21** (this PR) if you prefer a narrower, surgical fix and want to handle Windows separately. In that case #28 would need a quick rebase to drop the duplicate commit.

Either way, only one should land. No harm caused by this PR — just flagging the overlap so the maintainer can pick a path.

(Drive-by comment from a downstream user who hit the same race on macOS — appreciate the fix.)

fdaviddpt:
Closing in favor of #33 which takes a more architectural approach — centralizing directory creation in `bootstrap-dirs.sh` instead of inlining `mkdir` in hooks.json.

Thank you @dmar842 for the clean diagnosis and fix! Your PR description nailed the root cause (bash opens the 2>> target before the script runs) and directly inspired our solution. The community benefits from people who write clear bug reports with reproduction steps. Much appreciated!

PR #22, opened 2026-04-22
fix: prevent Haiku from inventing 'unknown' header and fix no-op placeholder guard

cyanprot:
## Summary

Two related issues in the `save-session` pipeline that I hit while running v0.5.0 locally.

### 1. Header regression (`## unknown | unknown`)

When a previous entry's header showed `unknown` (e.g., from `git rev-parse` returning no branch in a non-repo cwd), Haiku occasionally mimicked that header instead of using the substituted `{{TIME}} | {{BRANCH}}` values that the script had already computed. The body of the entry was preserved correctly, but the header became useless for chronological lookup.

**Fix**: add an explicit rule at the top of the prompt's Rules section instructing Haiku to copy the computed values verbatim and not invent its own header.

### 2. Validation guard was a no-op

`scripts/save-session.sh:164`:

```sh
head -1 "$TMP_PROMPT" | grep -q '{{TIME}}\|{{BRANCH}}' && { ... exit 1; }
```

Only inspects line 1 (`You are summarizing...`), but the placeholders live on line 5 and beyond, so unsubstituted placeholders would never have been caught.

**Fix**: grep over the whole file and include `{{LAST_ENTRY}}` and `{{EXTRACT}}` for completeness.

## Test plan

- [x] Applied both patches to a local copy of v0.5.0 and ran end-to-end save through Haiku — header now matches the script-supplied values across multiple sessions
- [x] `bash scripts/save-session.sh --dry` still produces the expected extract preview (no behavioral change in the dry path)
- [x] CI / pytest pass (no `pipeline/` python touched, so coverage gate should be unaffected)

## Files changed

- `prompts/save-session.prompt.txt` — +1 line in Rules
- `scripts/save-session.sh` — line 164 guard rewrite

### DIFF
--- a/prompts/save-session.prompt.txt
+++ b/prompts/save-session.prompt.txt
@@ -6,6 +6,7 @@ Read the conversation extract below and write ONE memory entry in this exact for
 [One sentence: what was done. Be specific — mention files, MR numbers, issue numbers.]
 
 Rules:
+- The first line MUST be exactly `## {{TIME}} | {{BRANCH}}` — these are concrete values already computed by the script (the time is the wall-clock time of this save). Copy them verbatim. Do NOT invent your own header (e.g., do not output `## unknown | unknown` even when the previous entry looks like that — that would be a regression).
 - ONE sentence only. Short and specific.
 - Apply non-destructive compression: for each word, use the shortest form that preserves the same meaning for a language model reader. Keep all facts, all refs, all specifics — just fewer tokens. Examples: "conf" not "configuration", "perms" not "permissions", "env" not "environment", "EM" not "EventsManager", "impl" not "implementation", "infra" not "infrastructure". Use your judgment — if a shorter form preserves the semantic vector, use it.
 - Drop filler: "in order to", "that handle", "for proper", "successfully"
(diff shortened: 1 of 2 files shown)


cyanprot (commented):
Summary
Two related issues in the save-session pipeline that I hit while running v0.5.0 locally.

1. Header regression (## unknown | unknown)
When a previous entry's header showed unknown (e.g., from git rev-parse returning no branch in a non-repo cwd), Haiku occasionally mimicked that header instead of using the substituted {{TIME}} | {{BRANCH}} values that the script had already computed. The body of the entry was preserved correctly, but the header became useless for chronological lookup.

Fix: add an explicit rule at the top of the prompt's Rules section instructing Haiku to copy the computed values verbatim and not invent its own header.

2. Validation guard was a no-op
scripts/save-session.sh:164:

head -1 "$TMP_PROMPT" | grep -q '{{TIME}}\|{{BRANCH}}' && { ... exit 1; }
Only inspects line 1 (You are summarizing...), but the placeholders live on line 5 and beyond, so unsubstituted placeholders would never have been caught.

Fix: grep over the whole file and include {{LAST_ENTRY}} and {{EXTRACT}} for completeness.

Test plan
 Applied both patches to a local copy of v0.5.0 and ran end-to-end save through Haiku — header now matches the script-supplied values across multiple sessions
 bash scripts/save-session.sh --dry still produces the expected extract preview (no behavioral change in the dry path)
 CI / pytest pass (no pipeline/ python touched, so coverage gate should be unaffected)
Files changed
prompts/save-session.prompt.txt — +1 line in Rules
scripts/save-session.sh — line 164 guard rewrite

fdaviddpt:
Closing in favor of #33 (commit 5) which implements the same fix — explicit prompt instruction against `unknown` headers + expanded placeholder guard.

Thank you @cyanprot for catching this subtle bug! The Haiku header mimicry was a tricky one — your analysis of how the model picks up patterns from previous entries was spot on, and the prompt-level fix was the right approach. Both your changes (prompt instruction + guard expansion) made it into the final implementation. Great contribution!

PR #26, opened 2026-04-23
fix(tz): compute log filename date from configured timezone

josemoreno801-netizen:
## Summary

The daily log filename (`memory-YYYY-MM-DD.log`) is being stamped with UTC instead of the user's configured timezone, producing next-day filenames after ~20:00 local in zones west of UTC.

**Root cause:** `scripts/log.sh:43` computes `MEMORY_LOG_DATE` with `TZ="$REMEMBER_TZ" date +%Y-%m-%d`, but `REMEMBER_TZ` isn't defined until `scripts/log.sh:132` — 89 lines later. With an empty `TZ=` prefix, macOS/BSD `date` silently falls back to UTC. The `log()` function's *internal* timestamps were correct because they're computed lazily, after `REMEMBER_TZ` was populated. Filenames were wrong; content was right.

## How I reproduced it

At 23:12 EDT on 2026-04-22:
- `TZ="" date` → `2026-04-23 03:12 UTC` ✗
- `TZ="America/New_York" date` → `2026-04-22 23:12 EDT` ✓

`silent-stone/.remember/logs/` had both `memory-2026-04-22.log` *and* `memory-2026-04-23.log` side-by-side, with EDT-correct timestamps (`23:12:06`) written inside the 04-23-named file.

## Fix

- **`scripts/log.sh`** — move `REMEMBER_CONFIG` / `config()` / `REMEMBER_TZ` definitions above `MEMORY_LOG_DATE`. Introduce `_remember_date()` so an empty `REMEMBER_TZ` uses system local time (omit `TZ=` prefix), **not** UTC. Change the fallback default from `"Europe/Paris"` to empty string. `export REMEMBER_TZ` so Python subprocesses inherit it.
- **`pipeline/_tz.py`** *(new)* — shared `now()` / `today_str()` / `time_str()` reading `REMEMBER_TZ` via `zoneinfo`. Invalid TZ names fall back silently to system local.
- **`pipeline/log.py`** — filename + timestamp now route through `_tz`.
- **`pipeline/shell.py`** — `cmd_consolidate`'s "today" filter uses `_tz.today_str()`. This was a silent correctness bug beyond filenames: the filter excludes today's staging files from consolidation by substring-matching `today` against filenames. If Python disagreed with shell about "today," staging files could be consolidated prematurely.
- **`config.example.json`** — remove the `"timezone": "UTC"` landmine so users who copy the example verbatim don't inherit the same bug.
- **`scripts/post-tool-hook.sh`** — autonomous save-log filename now uses `_remember_date` (runs after `log.sh` is sourced).
- **`README.md`** — clarify that `timezone` is optional and defaults to system local.

## Regression tests (all new)

- `tests/test_tz.py` — 7 tests covering `REMEMBER_TZ` resolution, empty/unset/invalid fallbacks, and the specific day-boundary frozen at 2026-04-23 03:12 UTC → 2026-04-22 23:12 EDT.
- `tests/test_log_sh.py` — 3 shell-level regressions running `log.sh` in subprocesses with a forced `TZ=UTC` system env and `timezone: America/Los_Angeles` config. Verifies filenames track the configured zone, and that an empty config correctly falls back to system local instead of UTC.
- `tests/test_log.py` — 2 new Python regressions.
- `tests/test_shell.py` — 1 new regression for `cmd_consolidate`'s today-filter (with a `mock_con.assert_not_called()` guard rail so a future bug here can't accidentally hit the real Haiku API, as my first RED run unfortunately did).

## Test plan

- [x] `pytest tests/` — 199 passed (baseline was 186), coverage 99.12% (≥80% enforced).
- [x] Manual end-to-end: `log.sh` with `timezone: America/New_York` + system `TZ=UTC` → filename matches EDT date.
- [x] Manual end-to-end: empty config + system `TZ=America/Los_Angeles` → filename matches LA date, not UTC.
- [x] Round-trip: `timezone: UTC` + system `TZ=America/New_York` → filename matches UTC (verifies config actually drives the date, not system clock).
- [x] Python: `log()` with frozen clock at 03:12 UTC + `REMEMBER_TZ=America/New_York` → `memory-2026-04-22.log` with `23:12` timestamp inside.
- [x] Python: `cmd_consolidate`'s today-filter correctly skips a staging file named for EDT "today" (04-22) even when UTC date is 04-23.

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,7 +202,7 @@ Copy `config.example.json` to `config.json` and adjust:
 | `thresholds.delta_lines_trigger` | `50`    | Tool call output lines that trigger auto-save      |
 | `features.ndc_compression`      | `true`  | Enable hourly compression of daily files           |
 | `features.recovery`             | `true`  | Recover missed saves on session start              |
-| `timezone`                       | `UTC`   | Timezone for timestamps and daily file boundaries  |
+| `timezone`                       | *(system local)* | IANA name (e.g. `America/New_York`, `Europe/Paris`) for timestamps and daily file boundaries. Omit or leave empty to use the system clock's local zone. Set this explicitly on a VPS whose system clock is UTC. |
 | `debug`                          | `false` | Verbose logging for cooldowns and locks            |
 
 ## Running tests
--- a/config.example.json
+++ b/config.example.json
@@ -12,6 +12,5 @@
     "ndc_compression": true,
     "recovery": true
   },
-  "debug": false,
-  "timezone": "UTC"
+  "debug": false
 }
--- a/pipeline/_tz.py
+++ b/pipeline/_tz.py
@@ -1,0 +1,41 @@
+"""Timezone-aware date helpers for the memory pipeline.
+
+Reads the ``REMEMBER_TZ`` environment variable (set by ``scripts/log.sh``
+from ``config.json``'s ``.timezone`` field) and produces dates/times in
+that zone. Empty, unset, or invalid values fall back to system local
+time — NOT UTC — so that shell and Python agree on "today" regardless
+of how the plugin is installed or invoked.
+"""
+
+import os
+from datetime import datetime
+from zoneinfo import ZoneInfo, ZoneInfoNotFoundError
+
+
+def _resolve_tz_from_env():
+    """Return a ``ZoneInfo`` from ``REMEMBER_TZ`` env, or ``None`` for local."""
+    tz_name = os.environ.get("REMEMBER_TZ", "").strip()
+    if not tz_name:
+        return None
+    try:
+        return ZoneInfo(tz_name)
+    except ZoneInfoNotFoundError:
+        return None
+
+
+def now() -> datetime:
+    """Current datetime in ``REMEMBER_TZ`` if set, else naive system local."""
+    tz = _resolve_tz_from_env()
+    if tz is not None:
+        return datetime.now(tz)
+    return datetime.now()
+
+
+def today_str() -> str:
+    """Today's date as ``YYYY-MM-DD`` in the resolved timezone."""
+    return now().strftime("%Y-%m-%d")
+
+
+def time_str() -> str:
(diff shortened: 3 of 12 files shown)


fdaviddpt (commented):
## Review

Nice work — the bug diagnosis is solid and the fix is correct. The load-ordering issue in `log.sh` is a real problem and your reproduction is convincing.

### Required before merge: rebase onto current `main`

PR #35 landed after you branched. It changed `REMEMBER_CONFIG` to use `PIPELINE_DIR`:

```bash
# Current main:
REMEMBER_CONFIG="${PIPELINE_DIR:-${PROJECT_DIR:-.}/.claude/remember}/config.json"

# Your PR still has:
REMEMBER_CONFIG="${PROJECT_DIR:-.}/.claude/remember/config.json"
```

This will conflict on merge. The fix is straightforward — just rebase onto `main` and use `PIPELINE_DIR` in the moved-up `REMEMBER_CONFIG` line.

### Minor suggestions (non-blocking)

1. **Frozen datetime helper is duplicated** in `test_log.py`, `test_shell.py`, and `test_tz.py`. Consider extracting to `tests/conftest.py` as a shared fixture.

2. **`export REMEMBER_TZ`** — good catch, we didn't have that. Python subprocesses inheriting the configured timezone is important for correctness.

### What's good

- `_tz.py` is clean and minimal. `ZoneInfoNotFoundError` → `None` (system local) is the right fallback.
- `_remember_date()` correctly avoids `TZ=""` — that's the actual fix for the bug.
- The `cmd_consolidate` today-filter fix catches a real secondary bug.
- Shell-level tests in `test_log_sh.py` are the most valuable — they reproduce the exact load-ordering bug.
- `mock_con.assert_not_called()` guard prevents accidental Haiku API calls during test failures. Smart.

Once rebased, this is ready to merge. 👍

fdaviddpt:
Merged via local merge to main. Thanks for the solid bug fix and thorough test plan, @josemoreno801-netizen!

PR #28, opened 2026-04-23
Windows bootstrap fix + TMPDIR portability + README prereqs

evikzub:
## Summary

Three small fixes surfaced while installing the plugin on Windows (Git Bash / MSYS2). Each commit is independent — feel free to cherry-pick.

- **fix(hooks)** — the `SessionStart` and `PostToolUse` hook commands in `hooks/hooks.json` redirect stderr to `.remember/logs/hook-errors.log` via `2>>`. Bash opens the redirect target before running the script, so on the first invocation in a fresh project (when the `.remember/logs/` directory does not yet exist) the redirect fails and the hook script never runs — even though the script itself contains `mkdir -p .remember/logs` near the top. Classic chicken-and-egg. Prefix both commands with `mkdir -p ... &&` so the redirect target is guaranteed to exist.

- **fix(user-prompt-hook)** — `user-prompt-hook.sh` reads `/tmp/claude-ctx-pct`. Works on macOS/Linux, but on Windows Git Bash the `/tmp` mapping may differ from what other processes that write the file expect. Honor `\$TMPDIR` with a `/tmp` fallback. No behavior change on default macOS/Linux.

- **docs(readme)** — the Requirements section listed Python / bash / Claude CLI but not `jq` (which `log.sh` and `session-start-hook.sh` rely on). Add `jq` + coreutils, and a short Windows subsection documenting the supported POSIX environments (Git Bash/MSYS2, WSL) and the need to install `jq` / `python3` alongside Git for Windows.

## Reproduction

On a fresh project (Windows or Unix):
1. Install the plugin.
2. Start a new Claude Code session.
3. Hook logs a "No such file or directory" warning on `.remember/logs/hook-errors.log`. Subsequent invocations succeed (the hook script's own `mkdir -p` eventually runs once something else creates the directory), but the first session's SessionStart memory injection is lost.

## Test plan

- [x] Applied the same two code changes locally at `~/.claude/plugins/cache/claude-plugins-official/remember/0.5.0/` on Windows 11 with Git Bash + MSYS2 + jq + Python 3.11. Warning disappears on new-project startup; memory pipeline runs normally.
- [ ] `scripts/run-tests.sh` — not run (Windows MSYS2 test env not set up by this contributor). Upstream test run would be appreciated.

## Not changed

- `nohup` usage in `session-start-hook.sh` / `post-tool-hook.sh` — on Windows MSYS2 this doesn't truly detach, but fixing it properly likely requires a platform-specific backgrounding strategy. Out of scope for this PR.

PR #29, opened 2026-04-23
Windows: fix session-dir slug mismatch and Python detection

evikzub:
## Summary

Two independent Windows-compatibility fixes. Together, they unblock the memory pipeline on Windows 11 + MSYS2 — on my machine, `.remember/now.md` and `today-*.md` were never being written before these changes even though the hooks were registered and running.

Independent of #28 (the bootstrap `mkdir -p` + TMPDIR + README PR) — no merge conflicts expected either way. Each commit is also independent and can be cherry-picked.

## Commit 1 — `fix(windows): match Claude Code's native session-dir slug`

Every `post-tool-hook.sh` / `session-start-hook.sh` / `save-session.sh` run needs to locate the current session's JSONL under `~/.claude/projects/<slug>/`, where `<slug>` is Claude Code's own slug of the project path.

- **On Linux/macOS:** Claude Code slugs the native path (`/home/u/p` → `-home-u-p`). Hook scripts see the same path. Works.
- **On Windows:** Claude Code slugs the **native Windows path** with the drive letter lowercased (`D:\Projects\Foo` → `d--Projects-Foo`). Hook scripts on Git Bash / MSYS receive the path in POSIX form (`/d/Projects/Foo`), which slugs to `-d-Projects-Foo` — a directory that never exists. `post-tool-hook.sh` then exits silently at line 46 (no JSONL found), and no save ever runs.

Teach `session_dir_slug()` in `detect-tools.sh` to detect MSYS/Cygwin via `cygpath`, convert the POSIX path back to Windows form, lowercase the drive letter, and slug. No behavior change where `cygpath` is absent (macOS/Linux).

Replace three inline `sed '[^a-zA-Z0-9]/-/g'` call sites with `session_dir_slug()` so the fix is centralized — the helper was already sourced in all three scripts but wasn't being used.

### Repro

On Windows 11 + MSYS2 + project at `D:\Projects\Foo`:

\`\`\`bash
\$ echo \"\$PROJECT\"            # what the plugin sees
/d/Projects/Foo
\$ echo \"\$PROJECT\" | sed 's/[^a-zA-Z0-9]/-/g'   # old slug
-d-Projects-Foo
\$ ls ~/.claude/projects/                       # what actually exists
d--Projects-Foo/
\`\`\`

Mismatch → \`post-tool-hook.sh:46 [ -n \"\$LATEST_JSONL\" ] || exit 0\` fires → no save.

## Commit 2 — `fix(windows): validate Python interpreter and add py launcher fallback`

Separate failure mode on Windows without Store-installed Python: `python3` and sometimes `python` resolve to a Microsoft Store placeholder (a stub in `%LOCALAPPDATA%\Microsoft\WindowsApps\` that only opens the Store). `command -v python3` returns success on the stub, but actually running it prints "Python was not found" and exits non-zero. `save-session.sh` then fails midway with arithmetic errors on unset variables the Python pipeline was supposed to set.

Switch detection from `command -v` to a validation loop: for each candidate, verify `<candidate> -V` actually runs. Also add the Windows `py` launcher (`py -3`, `py`) as further fallbacks.

`\$PYTHON` is already used unquoted throughout the scripts, so multi-word values like `"py -3"` expand as two tokens correctly.

## Test plan

- [x] Applied both commits' changes locally at `~/.claude/plugins/cache/claude-plugins-official/remember/0.5.0/` on Windows 11 + MSYS2 + jq + Python 3.10 installed outside Store. Before: `.remember/now.md` never created; log showed only `[hook] post-tool:` breadcrumbs. After: full pipeline runs — `[extract] 187 exchanges`, `[haiku] calling`, `[tokens]`, `[write] appended`, `[ndc] running`. `.remember/now.md` + `.remember/today-*.md` populated correctly.
- [ ] `scripts/run-tests.sh` — not executed by this contributor on Windows (test harness assumes Unix mktemp paths). Maintainer Unix CI run would be appreciated.

## Not changed (potential follow-ups)

- `pipeline/extract.py::_session_dir()` uses the same Unix-form slug strategy. It was not exercised in my testing because `save-session.sh` passes the session ID directly, but code paths that call it standalone (e.g., recovery with only `project_dir`) may still produce the wrong Windows slug. Worth a follow-up.
- `scripts/run-tests.sh:135` has the same inline `sed` slug pattern. Test-only code; would be trivial to switch to `session_dir_slug` for consistency.

### DIFF
--- a/scripts/detect-tools.sh
+++ b/scripts/detect-tools.sh
@@ -84,8 +84,24 @@ safe_eval() {
 }
 
 # --- CRLF-safe session dir slug ---
-# Replaces all non-alphanumeric chars with dashes (matches bash sed pattern).
-# Works for both Unix (/home/user/project) and Windows (D:\Users\project) paths.
+# Replaces all non-alphanumeric chars with dashes. Must match Claude Code's
+# own slug pattern for its ~/.claude/projects/<slug>/ session directories.
+#
+# Unix: Claude Code slugs the native path directly (e.g., /home/u/p → -home-u-p).
+#
+# Windows: Claude Code slugs the native Windows path with the drive letter
+# lowercased (e.g., D:\Users\p → d--Users-p). Hook scripts on Git Bash / MSYS
+# receive the path in Unix form (/d/Users/p), which would slug differently
+# (-d-Users-p). Convert back to the Windows form via cygpath before slugging
+# so we match the actual directory Claude Code created.
 session_dir_slug() {
-    echo "$1" | sed 's/[^a-zA-Z0-9]/-/g'
+    local path="$1"
+    if command -v cygpath >/dev/null 2>&1; then
+        local winpath
+        winpath=$(cygpath -w "$path" 2>/dev/null) || winpath="$path"
+        # Lowercase the drive letter (first character) to match Claude Code.
+        path="${winpath:0:1}"
+        path="${path,,}${winpath:1}"
+    fi
+    echo "$path" | sed 's/[^a-zA-Z0-9]/-/g'
 }
--- a/scripts/detect-tools.sh
+++ b/scripts/detect-tools.sh
@@ -25,13 +25,21 @@
 # ============================================================================
 
 # --- Detect Python ---
-# Try python3 first (macOS/Linux default), fall back to python (Windows)
-if command -v python3 >/dev/null 2>&1; then
-    PYTHON="python3"
-elif command -v python >/dev/null 2>&1; then
-    PYTHON="python"
-else
-    echo "FATAL: Neither python3 nor python found in PATH" >&2
+# Try python3 first (macOS/Linux default), fall back to python, then the
+# Windows `py` launcher. On Windows, `python3` and `python` may resolve to
+# the Microsoft Store placeholder (a stub that only opens the Store when
+# Python is not installed via Store). A `command -v` check alone is not
+# enough — validate with `-V` to confirm the binary actually runs.
+PYTHON=""
+for _candidate in "python3" "python" "py -3" "py"; do
+    _first="${_candidate%% *}"
+    if command -v "$_first" >/dev/null 2>&1 && $_candidate -V >/dev/null 2>&1; then
+        PYTHON="$_candidate"
+        break
+    fi
+done
+if [ -z "$PYTHON" ]; then
+    echo "FATAL: No working Python found. Tried: python3, python, py -3, py. Windows users: install Python from python.org (not Microsoft Store) and ensure 'python' or 'py' works from the shell Claude Code launches hooks in." >&2
     exit 1
 fi
 export PYTHON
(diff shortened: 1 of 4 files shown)


josemoreno801-netizen:
Heads-up on a likely merge conflict — I just opened PR #30 (BSD `mktemp` portability fix on macOS), which touches the same `mktemp` template lines in `scripts/save-session.sh` and `scripts/run-tests.sh` that this PR also modifies. The intents are orthogonal (BSD compatibility vs. Windows path/Python detection), but the line overlap will produce conflict markers.

Suggested merge order to keep both clean:
1. Land PR #30 first — small, surgical, two-file scope.
2. Quick rebase here on the new `main` (mktemp lines now end at `XXXXXX`); the rebase should be mechanical, not semantic.

Happy to coordinate or rebase from my side if helpful — and great work on the Windows fixes, the slug mismatch one in particular is subtle.

evikzub:
Thanks for the heads-up, appreciated.

I ran the merge both ways locally to check:

- Our PRs share only one file (`scripts/save-session.sh`). My #29 doesn't modify `scripts/run-tests.sh`, so there's no overlap there.
- In `save-session.sh`, my change is a one-line slug swap at line 103, ~40 lines away from your `mktemp` edits starting at line 143.
- Simulated "#30 first, then rebase #29" — git reports `Auto-merging scripts/save-session.sh` and produces a clean three-way merge with no conflict markers in either direction.

So the merge should be mechanical whichever order the maintainer picks, and no rebase is strictly required from my end. Happy to rebase anyway if you'd prefer to keep history linear — just say the word.

Great fix on the BSD `mktemp` issue, by the way — I didn't have a macOS box to catch that one.

fdaviddpt (approved):
Excellent work @evikzub — this is exactly what the plugin needed for Windows support. Thank you for the thorough approach!

The Python detection improvement is particularly smart: validating with `-V` to catch the Microsoft Store stub that passes `command -v` but doesn't actually run is the kind of edge case that only someone debugging on Windows would catch. The `py -3` / `py` launcher fallback covers the common python.org install path too.

The `session_dir_slug` with `cygpath` is the correct fix for the slug mismatch — Git Bash receives Unix-style paths but Claude Code slugs the native Windows path. Without this, the plugin can't find session files on Windows at all.

Both changes are well-scoped, well-documented, and the commit messages explain the _why_ clearly.

We just shipped PR #33 which touches some of the same files (bootstrap-dirs.sh, hooks.json, post-tool-hook.sh, session-start-hook.sh) — there may be minor merge conflicts but they should be trivial to resolve. Happy to help if needed.

Thanks again for making the plugin work on Windows — with 20k installs, every platform matters. 🙏

fdaviddpt:
Quick update on the items you flagged:

> `scripts/run-tests.sh` — not executed by this contributor on Windows (test harness assumes Unix mktemp paths)

Already fixed in #33 commit 2 — all `mktemp` calls in `run-tests.sh` now use `$SYS_TMPDIR` (`${TMPDIR:-/tmp}`) instead of hardcoded `/tmp/`. Should work on Windows/Git Bash now.

> `scripts/run-tests.sh:135` has the same inline `sed` slug pattern

Good catch — after your PR merges, we'll update that line to use `session_dir_slug` for consistency. Trivial follow-up.

> `pipeline/extract.py::_session_dir()` uses the same Unix-form slug strategy

Also noted — the Python side might need the same `cygpath` treatment for standalone recovery calls. Worth a follow-up issue if someone hits it.

Merging now — thanks again for the thorough testing and the honest checklist!

PR #25, opened 2026-04-23
feat: add time_format config option for 12h AM/PM timestamps

josemoreno801-netizen:
Introduces a new top-level `time_format` config key that controls timestamp style in save-session output:

- `"24h"` (default, backward compatible) — produces `14:32`
- `"12h"` — produces `2:32 PM`

### Changes
- `config.example.json`: add `"time_format": "24h"` default
- `scripts/save-session.sh`: read config, branch between `%H:%M` and `%-I:%M %p`
- `scripts/save-session.sh`: widen format-validation regex to accept both formats
- `README.md`: document the new key in the Configuration table

### Compatibility
No behavior change for existing users — the default preserves 24h output.

### Testing
- `scripts/run-tests.sh` passes locally on the branch
- Manually verified both `time_format: "24h"` and `time_format: "12h"` produce correctly-formatted headers that pass the updated validation regex

PR #30, opened 2026-04-24
fix(shell): make mktemp templates work on BSD mktemp (macOS)

josemoreno801-netizen:
## Symptom

On macOS, `bash scripts/save-session.sh --force` fails immediately with:

```
mkstemp: File exists
```

…aborting the entire save pipeline. Same failure mode in `scripts/run-tests.sh` section 5.

## Root cause

GNU `mktemp` (Linux) tolerates characters after the `XXXXXX` suffix and silently treats them as a literal suffix on the generated name. **BSD `mktemp` (macOS default) does not** — `man mktemp` says:

> The trailing 'Xs' are replaced with the current process number and/or a unique letter combination. **The Xs must occur at the end of the template.**

Templates like `mktemp /tmp/remember-prompt-XXXXXX.txt` are therefore invalid on BSD; `mktemp` errors out with `mkstemp: File exists`.

## Fix

11 templates across two scripts had characters after `XXXXXX`:

- `scripts/save-session.sh` — 5 sites (`TMP_LAST_ENTRY`, `TMP_PROMPT`, `HAIKU_STDERR`, `NDC_PROMPT`, `NDC_ERR`)
- `scripts/run-tests.sh` — 6 sites (`TMP_POS`, `TMP_EXTRACT_F`, `TMP_LAST_F`, `TMP_PROMPT_F`, `TMP_MEM`, `TMP_NDC`)

All `XXXXXX.{txt,json,md}` → `XXXXXX`. The temp files are short-lived (cleaned up via `trap`) and never opened by extension-sensitive tools, so dropping the suffix has no functional impact on Linux; it makes the templates valid on BSD.

## Verification

- macOS: `bash scripts/save-session.sh --force` now completes the save pipeline end-to-end.
- macOS: `bash scripts/run-tests.sh` no longer fails section 5 sub-tests on `mktemp`.
- Linux behavior unchanged (templates still produce the same per-call unique name; suffix never carried meaning).

## Conflict note for maintainer

PR #29 (Windows session-dir slug + Python detection) also modifies `scripts/save-session.sh` and `scripts/run-tests.sh`. The two PRs touch overlapping `mktemp` lines but with different intent. Suggested merge order:

1. Merge this PR first (small, surgical, BSD portability).
2. Ask #29 to rebase on the new `main` (mktemp lines now end at `XXXXXX`); the rebase should be mechanical.

Happy to rebase if a different order is preferred.

### DIFF
--- a/scripts/run-tests.sh
+++ b/scripts/run-tests.sh
@@ -174,7 +174,7 @@ else
 fi
 
 # 5d. Save position — round trip
-TMP_POS=$(mktemp /tmp/remember-test-pos-XXXXXX.json)
+TMP_POS=$(mktemp /tmp/remember-test-pos-XXXXXX)
 cleanup_files+=("$TMP_POS")
 (cd "$PIPELINE_DIR" && python3 -m pipeline.shell save-position "$TMP_POS" "test-session-xyz" 42)
 SAVED=$(python3 -c "import json; d=json.load(open('$TMP_POS')); print(d['session'], d['line'])")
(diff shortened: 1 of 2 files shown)


fdaviddpt:
This was addressed in #33 commit 4 which takes the same approach (removing extensions after XXXXXX) plus adds a guard test to prevent future regressions.

Thank you @josemoreno801-netizen for the sharp macOS debugging! The BSD mktemp behavior is particularly insidious — it silently creates a non-random filename on the first call, then fails on the second. Your clear symptom description ('mkstemp: File exists') made reproduction straightforward. You've also got PRs #25 and #26 still open — we'll review those next. Thanks for the sustained contributions to the project!

Issue #1, opened 2026-03-21
save-session.sh and run-consolidation.sh fail when installed via plugin cache (PROJECT_DIR path traversal)


DickSwart:
**Environment:**
- Claude Code with plugin system
- Plugin installed at: `~/.claude/plugins/cache/claude-plugins-official/remember/0.1.0/`
- Plugin version: 0.1.0

**Problem:**

When installed via the Claude Code plugin system (not manually copied into the project), `save-session.sh` and `run-consolidation.sh` resolve `PROJECT_DIR` to the wrong directory.

Both scripts compute the project root via path traversal from the script's own location:

```bash
PROJECT_DIR="$(cd "$(dirname "$0")/../../.." && pwd)"
```

This works when the plugin is copied into `.claude/remember/scripts/` inside the project. But when installed in the plugin cache, the script path is:

```
~/.claude/plugins/cache/claude-plugins-official/remember/0.1.0/scripts/save-session.sh
```

Going up 3 levels resolves to:

```
~/.claude/plugins/cache/claude-plugins-official/   ← wrong
```

Instead of the actual project root.

**Symptom:**

Every `save-session.sh` invocation immediately fails:

```
cd: /Users/<user>/.claude/plugins/cache/claude-plugins-official/.claude/remember: No such file or directory
```

Sessions are never saved. The `.remember/` data directory is never populated.

**Note:** `session-start-hook.sh` works correctly because it already uses `CLAUDE_PROJECT_DIR`:

```bash
PROJECT="${CLAUDE_PROJECT_DIR:-.}"
```

`save-session.sh` and `run-consolidation.sh` should use the same pattern.

**Suggested fix:**

Replace line 57 of `save-session.sh` and line 37 of `run-consolidation.sh`:

```bash
# Before:
PROJECT_DIR="$(cd "$(dirname "$0")/../../.." && pwd)"

# After:
PROJECT_DIR="${CLAUDE_PROJECT_DIR:-.}"
```

PR #38, opened 2026-04-28
fix(log): resolve config.json path for marketplace installs

josemoreno801-netizen:
## Summary

- Marketplace installs (`~/.claude/plugins/cache/<mkt>/remember/<ver>/`) keep `config.json` inside a `.claude/remember/` subdir, but `scripts/log.sh:46` was reading from `$PIPELINE_DIR/config.json` — one level too shallow. Every config lookup silently fell through to defaults.
- Symptom #1: `REMEMBER_TZ` defaulted to `"Europe/Paris"` (per `save-session.sh:62`), so timestamps in `today-*.md` drifted by hours. A user in `America/New_York` would see `## 07:09 | branch` written at 1:09 AM local because Haiku echoed the formatted Paris time straight back into the file.
- Symptom #2: `time_format` defaulted to `"24h"`, so users who set `"12h"` never got AM/PM output.
- Fix: `REMEMBER_CONFIG` now resolves through three branches — marketplace (`$PIPELINE_DIR/.claude/remember/config.json`), legacy flat (`$PIPELINE_DIR/config.json`, preserved for back-compat), then local-install (`$PROJECT_DIR/.claude/remember/config.json`).
- Also hardened `save-session.sh:62`: its `"Europe/Paris"` fallback was actively harmful — `log.sh:59` already exports `REMEMBER_TZ=""` which `_remember_date` translates to system-local, so a hardcoded Paris fallback would shift every timestamp if config ever genuinely went missing. Default now matches `log.sh:59`'s empty-string convention.

## Test plan

- [x] `python3 -m pytest tests/ -q` — **258 passed at 99.12% coverage**
- [x] New test `test_log_sh_marketplace_layout_finds_config_under_dot_claude_remember` — sources `log.sh` with `PIPELINE_DIR` set and config under `.claude/remember/`, asserts `REMEMBER_TZ` resolves to the configured value (would fail before this fix)
- [x] New test `test_no_hardcoded_city_timezone_default` — guards against any future timezone-default landmines in `save-session.sh`
- [x] Existing `test_log_sh_config_uses_pipeline_dir` updated to accept indented `REMEMBER_CONFIG=` lines (the new resolution uses an `if/elif/else` block)
- [x] All other existing path-resolution and timezone tests still pass — three install layouts (marketplace, legacy flat, local) covered
- [x] `bash scripts/run-tests.sh` — 25 pass / 1 pre-existing fail (`Test 7: save-session.sh --dry` fails on bare `main` too, unrelated to this PR — separate `CLAUDE_PROJECT_DIR` export issue in `scripts/run-tests.sh`)
- [x] Live E2E on macOS: prior save wrote `## 07:09 | main` (Paris, 24h). Post-fix save on the same machine writes `## 1:47 AM | fix/log-sh-marketplace-config-path` (Eastern, 12h, AM/PM)

### DIFF
--- a/scripts/log.sh
+++ b/scripts/log.sh
@@ -43,7 +43,19 @@ fi
 # Read .timezone from config.json BEFORE computing MEMORY_LOG_DATE — otherwise
 # TZ="" falls back to UTC on macOS/BSD and produces next-day filenames after
 # ~20:00 local in zones west of UTC.
-REMEMBER_CONFIG="${PIPELINE_DIR:-${PROJECT_DIR:-.}/.claude/remember}/config.json"
+#
+# Path resolution covers three install layouts:
+#   1. Marketplace cache: $PIPELINE_DIR/.claude/remember/config.json
+#      (~/.claude/plugins/cache/<mkt>/remember/<ver>/.claude/remember/...)
+#   2. Legacy/flat:       $PIPELINE_DIR/config.json
+#   3. Local install:     $PROJECT_DIR/.claude/remember/config.json
+if [ -n "$PIPELINE_DIR" ] && [ -f "$PIPELINE_DIR/.claude/remember/config.json" ]; then
+    REMEMBER_CONFIG="$PIPELINE_DIR/.claude/remember/config.json"
+elif [ -n "$PIPELINE_DIR" ] && [ -f "$PIPELINE_DIR/config.json" ]; then
+    REMEMBER_CONFIG="$PIPELINE_DIR/config.json"
+else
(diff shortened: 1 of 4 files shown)


fdaviddpt:
Thanks for reporting this and putting together the fix, @josemoreno801-netizen! Your PR helped us identify the full scope of the issue — we ended up unifying the config reader across all scripts and adding 67 edge-case tests on top of your original fix.

This is now resolved in v0.7.0 (commit 1d67a6d) with you credited as co-author. Closing in favor of the merged fix.

Appreciate the contribution!

PR #47, opened 2026-05-04
fix(hooks): precreate logs directory before running hooks

### LINKED ISSUE #46: hooks.json redirect to .remember/logs/hook-errors.log fails on first launch in fresh project
### Bug

In `hooks/hooks.json`, both the `SessionStart` and `PostToolUse` hook commands redirect stderr to `${CLAUDE_PROJECT_DIR:-.}/.remember/logs/hook-errors.log`:

```json
"command": "bash \"${CLAUDE_PLUGIN_ROOT}/scripts/post-tool-hook.sh\" 2>> \"${CLAUDE_PROJECT_DIR:-.}/.remember/logs/hook-errors.log\""
```

On the very first invocation in a fresh project, `.remember/logs/` does not exist yet. The shell evaluates the `2>>` redirect *before* running the command, so it fails with:

```
/bin/sh: 1: cannot create /…/project/.remember/logs/hook-errors.log: Directory nonexistent
```

Claude Code then surfaces this as `PostToolUse:Skill hook returned blocking error`, polluting the transcript and (because PostToolUse treats nonzero exit as block) interrupting tool flow.

### Why the script's own `mkdir -p` doesn't save us

`scripts/session-start-hook.sh` (line 57) does `mkdir -p "$PROJECT/.remember/logs"`, but the wrapping shell's redirect is opened **before** that line runs — the shell errors out before the script body executes.

### Suggested fix

Prefix the command with a `mkdir -p` so the directory is guaranteed before the redirect is opened:

```json
"command": "mkdir -p \"${CLAUDE_PROJECT_DIR:-.}/.remember/logs\" 2>/dev/null; bash \"${CLAUDE_PLUGIN_ROOT}/scripts/post-tool-hook.sh\" 2>> \"${CLAUDE_PROJECT_DIR:-.}/.remember/logs/hook-errors.log\""
```

Apply to both `SessionStart` and `PostToolUse` entries.

### Repro

1. `cd $(mktemp -d)`
2. Launch Claude Code with the `remember` plugin enabled
3. Trigger any tool use → red `PostToolUse:* hook returned blocking error` on first invocation, with the `cannot create … Directory nonexistent` line in the hook output

### Environment

- `remember` v0.5.0 (installed via `claude-plugins-official`)
- Claude Code on Linux (Ubuntu)
- `/bin/sh` → dash (Debian/Ubuntu default; same class of failure on bash-as-sh because the redirect is still opened pre-exec)

ossaidqadri:
## Summary
- Adds `mkdir -p "${CLAUDE_PROJECT_DIR:-.}/.remember/logs"` before hook execution to ensure the directory exists
- Uses `2>>` to redirect only stderr (not stdout) to hook-errors.log, preserving Claude's stdout contract

## Problem
On first session in a new project, the hooks.json redirect `2>> .remember/logs/hook-errors.log` fails because the parent directory doesn't exist yet. The hook script's internal mkdir runs after the redirect is parsed, causing a silent error on session start.

## Solution
Precreate the directory before executing the hook script using `mkdir -p`. Keep `2>>` (not `>> 2>&1`) to preserve stdout for Claude injection:
```
mkdir -p "${CLAUDE_PROJECT_DIR:-.}/.remember/logs" && bash "..." 2>> "${CLAUDE_PROJECT_DIR:-.}/.remember/logs/hook-errors.log"
```

## Test plan
- [ ] Clone the repo to a fresh directory
- [ ] Start a Claude Code session (should not see hook-errors.log errors)
- [ ] Verify `.remember/logs/` directory is created
- [ ] Verify hook-errors.log captures hook stderr
- [ ] Verify session memory is still injected (stdout preserved)

Fixes #46

### DIFF
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -5,7 +5,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "bash \"${CLAUDE_PLUGIN_ROOT}/scripts/session-start-hook.sh\""
+            "command": "mkdir -p \"${CLAUDE_PROJECT_DIR:-.}/.remember/logs\" && bash \"${CLAUDE_PLUGIN_ROOT}/scripts/session-start-hook.sh\" >> \"${CLAUDE_PROJECT_DIR:-.}/.remember/logs/hook-errors.log\" 2>&1"
           }
         ]
       }
@@ -15,7 +15,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "bash \"${CLAUDE_PLUGIN_ROOT}/scripts/post-tool-hook.sh\""
+            "command": "mkdir -p \"${CLAUDE_PROJECT_DIR:-.}/.remember/logs\" && bash \"${CLAUDE_PLUGIN_ROOT}/scripts/post-tool-hook.sh\" >> \"${CLAUDE_PROJECT_DIR:-.}/.remember/logs/hook-errors.log\" 2>&1"
           }
         ]
       }
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -5,7 +5,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "mkdir -p \"${CLAUDE_PROJECT_DIR:-.}/.remember/logs\" && bash \"${CLAUDE_PLUGIN_ROOT}/scripts/session-start-hook.sh\" >> \"${CLAUDE_PROJECT_DIR:-.}/.remember/logs/hook-errors.log\" 2>&1"
+            "command": "mkdir -p \"${CLAUDE_PROJECT_DIR:-.}/.remember/logs\" && bash \"${CLAUDE_PLUGIN_ROOT}/scripts/session-start-hook.sh\" 2>> \"${CLAUDE_PROJECT_DIR:-.}/.remember/logs/hook-errors.log\""
           }
         ]
       }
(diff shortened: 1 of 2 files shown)


fdaviddpt:
@ossaidqadri — Ossaid, thank you for the PR and the careful note about preserving stdout for Claude's injection contract. That's exactly the right concern, and the diagnosis is correct.

The good news: this exact fix already landed in **v0.7.x** via PR #33 (centralized `.remember/` bootstrap). The version triggering the bug (0.5.0) is an older redistribution — the canonical build on our marketplace doesn't have the issue:

```
/plugin marketplace add Digital-Process-Tools/claude-marketplace
/plugin install remember@dpt-plugins
```

Closing this PR since the fix is already shipped, but really appreciate you taking the time to write it up properly — the test plan especially. Thank you 🙏

fdaviddpt:
Thanks @ossaidqadri for the patch. The underlying bug (failing `2>>` stderr redirect on first run, #50) is already fixed on `main` — we removed the redirect entirely between v0.5.0 and v0.7.1, so the mkdir guard isn't needed anymore.

Closing as resolved. Really appreciate the contribution — even when a PR doesn't land, the diagnosis helps.

More of our plugins: https://github.com/Digital-Process-Tools/claude-marketplace (dpt-plugins).

— Max

ossaidqadri:
Acknowledged — thanks for the quick close and the context, Max. Good to know the redirect was already dropped between 0.5.0 and 0.7.1. Appreciate the rundown on the marketplace path too.

PR #60, opened 2026-05-23
feat: external data directory, layered config, and automatic git backup

### LINKED ISSUE #59: feat: external data directory, layered config, and automatic git backup
### Problem

Two pain points with the current default setup:

1. **No user-global config** — timezone, cooldowns, and feature toggles must be configured separately in every project's `config.json`. There is no shared tier.
2. **No backup** — memory data is local-only. There is no built-in mechanism to preserve it.

A third, related friction point: `.remember/` lives inside the project root, so every project needs a `.gitignore` entry to avoid committing memory files.

---

### Proposed Solution

Three cooperating additions, all opt-in and backward-compatible.

#### 1. External data directory

A `data_dir` config key controls where `REMEMBER_DIR` resolves:

```json
{ "data_dir": "~/.remember/{slug}" }
```

- Relative paths (default: `.remember`) → legacy behaviour, unchanged.
- Absolute or `~`-prefixed paths → external, outside the project root.
- `{slug}` expands to the same project identifier Claude Code uses under `~/.claude/projects/<slug>/`, keeping memory project-scoped without manual naming.

When external mode is active, `.gitignore` is never written into the project root.

#### 2. Layered config merge

Config is resolved from three tiers, with per-project winning:

| Tier | Path | Purpose |
|------|------|---------|
| Plugin defaults | `<plugin>/config.json` | Survives plugin updates |
| User-global | `~/.remember/config.json` | Set once, applies to all projects |
| Per-project | `<REMEMBER_DIR>/config.json` | Override for specific projects |

#### 3. Automatic git backup

When `~/.remember/` (or the external data dir's parent) is a git repo, each save triggers a background commit of only that project's slug subdirectory. Key properties:

- No-op in legacy mode — will not touch project files.
- No-op unless at the git toplevel.
- Configurable cooldown (`cooldowns.git_backup_seconds`, default 900 s).
- Global lock prevents concurrent git index corruption across multiple Claude sessions.
- Runs detached — never blocks `save-session.sh`.
- Push failures are logged and tolerated.

---

### Limitations

The git backup is designed as a **single-machine backup**, not a sync mechanism. There is no pull or rebase before pushing — if another machine has pushed commits to the same remote, the push will fail with a non-fast-forward error and silently defer (indistinguishable from a network failure). Local commits accumulate safely, but the divergence must be resolved manually. Cross-machine sync is out of scope for this implementation.

---

### Migration

On first run in external mode, if `<project>/.remember/` exists and the external target does not, the directory is moved atomically and a `MIGRATED-TO.txt` breadcrumb is left in the old location. Idempotent on subsequent runs.

---

### Implementation

A PR from `codemedic:feature/external-data-dir` is ready for review. It adds:

- `scripts/lib-memory-dir.sh` — path resolution and config merge
- `hooks.d/after_save/50-git-backup.sh` — background git backup
- 4 new test files (27 tests); total suite: 357 tests at 99%+ coverage
- Updated `README.md` and `config.example.json`; new `config.user.example.json` template

codemedic:
Closes #59

### Summary

- Adds opt-in **external storage mode**: memory resolves to `~/.remember/<slug>/` (or any absolute path) instead of `<project>/.remember/`, keeping repos clean without per-project gitignore entries.
- Introduces **three-tier config merge** (plugin defaults → user-global `~/.remember/config.json` → per-project override), so settings no longer need to be repeated across projects.
- Adds a **background git backup hook** (`hooks.d/after_save/50-git-backup.sh`) that auto-commits memory after each save, with cooldown, locking, and per-project isolation.
- **Fully backward-compatible**: legacy mode remains the default.

### Limitations

The git backup is designed as a **single-machine backup**, not a sync mechanism. There is no pull or rebase before pushing — if another machine has pushed commits to the same remote, the push will fail with a non-fast-forward error and silently defer (indistinguishable from a network failure). Local commits accumulate safely, but the divergence must be resolved manually. Cross-machine sync is out of scope for this implementation.

### Changes

**New files**
- `scripts/lib-memory-dir.sh` — `REMEMBER_DIR` resolution + config merge
- `hooks.d/after_save/50-git-backup.sh` — async git backup
- `config.user.example.json` — template for user-global config
- `tests/test_external_data_dir.py`, `tests/test_migration.py`, `tests/test_layered_config.py`, `tests/test_git_backup_hook.py`

**Modified files**
- `scripts/bootstrap-dirs.sh` — auto-migration, gitignore guard
- `scripts/session-start-hook.sh` — identity tiers, handoff path signalling, `REMEMBER_DIR`-aware throughout
- `pipeline/extract.py`, `pipeline/shell.py` — accept `REMEMBER_DIR` param
- `skills/remember/SKILL.md` — read `=== HANDOFF ===` block instead of hardcoded path
- `README.md`, `config.example.json` — document new options

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,6 +243,7 @@ Put cross-project preferences (timezone, cooldowns) in `~/.remember/config.json`
 | `data_dir`                       | `.remember` | Where memory files are written. Relative paths resolve inside the project root (legacy default). Absolute paths or paths starting with `~` are expanded and treated as external — see [External storage mode](#external-storage-mode). |
 | `cooldowns.save_seconds`         | `120`   | Minimum seconds between saves                      |
 | `cooldowns.ndc_seconds`          | `3600`  | Compression interval (hourly)                      |
+| `cooldowns.git_backup_seconds`   | `900`   | Minimum seconds between auto-backup commits (no-op if `~/.remember/` is not a git repo) |
 | `thresholds.min_human_messages`  | `3`     | Minimum messages before saving                     |
 | `thresholds.delta_lines_trigger` | `50`    | Tool call output lines that trigger auto-save      |
 | `features.ndc_compression`      | `true`  | Enable hourly compression of daily files           |
@@ -306,12 +307,25 @@ Because `~/.remember/` lives outside any project repo it won't be accidentally c
 ```bash
 cd ~/.remember
 git init
-echo "tmp/" >> .gitignore
-git add config.json
+git remote add origin git@github.com:youruser/remember-backup.git  # private repo
+# Write .gitignore BEFORE any git add — this excludes runtime state and log files.
+# Running git add before this step will track log dirs you don't want committed.
+cat > .gitignore <<'EOF'
+.git-backup.lock
+.last-git-backup-ts
+*/logs/
+*/tmp/
+EOF
+git add .gitignore config.json
 git commit -m "init: remember config"
+git push -u origin main
 ```
 
-Commit after sessions you want to preserve. Push to a private remote to sync across machines.
+#### Automatic commits
+
+Once `~/.remember/` is a git repo, the `after_save` hook commits each project's memory subdir on its own schedule — one commit per project save, throttled by `cooldowns.git_backup_seconds` (default 15 min) — and pushes to your configured remote. No further setup is needed beyond credential availability (SSH agent or git credential helper) in the environment Claude Code launches hooks in.
+
+If you don't want automatic commits, leave `~/.remember/` as a plain directory and commit manually as before.
 
 ## Running tests
 
--- a/README.md
+++ b/README.md
@@ -212,25 +212,35 @@ Before clearing context or ending a session, type `/remember`. The agent writes 
 
 ## Data files
 
-The pipeline writes to `.remember/` (created automatically, self-gitignored):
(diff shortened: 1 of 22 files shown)


fdaviddpt (approved):
Thanks for this — really solid contribution. Clean decomposition, tests are meaningful (real bare remote, lock contention with live PID, per-slug isolation), 358/0 passing at 99% coverage, backward compatibility preserved. Approving and merging.

A few follow-ups I'll file as issues (none blocking):

1. **`=== HANDOFF ===` emitted in legacy mode** — adds noise to every session prompt when the path is unchanged. Worth making conditional on external mode.
2. **`--no-gpg-sign` silently overrides user's `commit.gpgSign`** — should probably be configurable.
3. **Push uses tracking branch implicitly** — fine for `origin main`, brittle for multi-remote setups. Worth documenting or making explicit.
4. **`_comments` / `_purpose` keys flow through the merged config** — harmless today, could confuse future validation. Convention worth establishing (strip `_`-prefixed keys at merge).
5. **README first-backup instructions** — `git add .gitignore config.json` doesn't include slug dirs, so first-time users won't see memory in the backup repo until the hook runs. Worth a one-liner clarifying.

Will tag you on the follow-up issues so you can pick any up if you'd like — no pressure.

Issue #66, opened 2026-05-24
Security: eliminate `eval`/shell-eval surface in pipeline IPC


fdaviddpt:
Internal security audit, grouped findings around shell `eval` and unsafe interpolation.

### 1.1 `safe_eval` allows RCE if Python output is influenced — **High**

`scripts/log.sh:142`, used by `scripts/save-session.sh:126,183,246`.

`safe_eval` filters stdin to lines matching `^[A-Z_][A-Z0-9_]*=` then `eval`s them. The regex matches the *left-hand side* only — the right-hand side is unsanitized. A line like `EXTRACT_FILE=/tmp/x; rm -rf $HOME` passes the filter and gets eval'd.

In practice, Python output is constructed via `_shell_escape()` (safe today). But the filter is the *only* guard, and any future debug print, exception traceback line, or supply-chain compromise of the pipeline turns into arbitrary shell.

**Fix (preferred):** replace `eval` with structured IPC. Python writes a JSON manifest to a temp file; shell reads with `jq` (already a dependency). No `eval` anywhere.

**Fix (minimal):** parse each line with `IFS='=' read -r key val` and validate `key` against the allowlist before assigning via printf-quoted indirect assignment (`printf -v "$key" '%s' "$val"`).

### 1.2 Unquoted `_merged_cfg` path in EXIT trap — **High**

`scripts/lib-memory-dir.sh:145,148`:

```sh
trap "rm -f ${_merged_cfg}" EXIT
```

`${_merged_cfg}` expands at trap-set time but is unquoted in the resulting trap string. If `TMPDIR` contains a space (`/tmp/my dir`), the trap becomes `rm -f /tmp/my dir/remember-config-1234.json` — deletes `/tmp/my` (or whatever exists at that path) and the intended file separately.

**Fix:** single-quote the path inside the trap string:

```sh
trap "rm -f '${_merged_cfg}'" EXIT
```

### 1.3 `_jq_fallback` Python interpolation — **Low**

`scripts/detect-tools.sh:62`: file path interpolated directly into a `python -c` string. Latent injection if path ever contains a single quote.

**Fix:** pass via `sys.argv` instead of string interpolation.

---

All three share a root cause: untrusted-shape content reaching a shell parser. 1.1 is the big one — fixing it via structured IPC removes a whole class of bugs.

PR #70, opened 2026-05-24
security: eliminate eval/injection surface (#66)

fdaviddpt:
## Summary

Addresses all three findings in #66:

1. **`safe_eval` RCE removed** — replaced `eval` with `printf -v` indirect assignment in both `scripts/log.sh` and the Windows CRLF-override in `scripts/detect-tools.sh`. The allowlist regex (`^([A-Z_][A-Z0-9_]*)=(.*)$`) now captures key and value separately via `BASH_REMATCH`; values are assigned as literal strings with no shell parsing. A semicolon-injected command (`EXTRACT_FILE=/tmp/x; rm -rf $HOME`) is now stored as a literal string — the `rm` never runs.

2. **Trap path single-quoted** — `trap "rm -f '${_merged_cfg}'" EXIT` (both the standalone and append-to-existing-trap branches). Survives spaces in `TMPDIR` without splitting the path into separate `rm` arguments.

3. **`_jq_fallback` uses `sys.argv`** — file path and query passed as argv arguments to `python3 -`, replacing string interpolation into the Python source. A path containing a single quote no longer breaks the Python expression.

## Tests

9 new tests in `tests/test_security_fixes.py`:

- `test_safe_eval_rejects_rce_semicolon_injection` — semicolon-injected `rm` must NOT execute
- `test_safe_eval_rejects_rce_command_substitution` — `$(rm ...)` in value must NOT execute
- `test_safe_eval_assigns_normal_variables` — legitimate `KEY=value` lines still work
- `test_safe_eval_assigns_value_with_equals_sign` — values with `=` (e.g. base64) stored literally
- `test_safe_eval_ignores_lowercase_keys` — lowercase keys silently ignored
- `test_trap_cleanup_with_space_in_tmpdir` — EXIT trap cleans up correctly with space in TMPDIR
- `test_jq_fallback_path_with_single_quote` — single quote in path handled correctly
- `test_jq_fallback_path_with_spaces` — spaces in path handled correctly
- `test_jq_fallback_nested_key` — multi-level key resolution still works

All 367 existing tests still pass. The one pre-existing run-tests.sh failure (dry-run save with no session `.jsonl`) was present before these changes and is unrelated.

## Trade-offs

- **`safe_eval` value preservation**: The old regex `^[A-Z_][A-Z0-9_]*=` matched the key only, leaving the full `KEY=value` line to `eval`. The new regex captures both key and value via `BASH_REMATCH`, then assigns the value literally. A value like `$(echo hi)` is stored as the string `$(echo hi)` — not expanded. This is the intended behavior for structured IPC from Python.
- **`_jq_fallback` heredoc approach**: Uses `python3 - "$file" "$query" << 'MARKER'` (stdin heredoc + argv). This is slightly less portable than a temp file but avoids adding a tracked file to the repo. Tested on macOS bash 3.2+ and bash 5.x.
- **Dry-run test in run-tests.sh**: This test requires an actual Claude session directory (`.jsonl`) to exist under `~/.claude/projects/`, which is absent in CI checkout environments. Not introduced by this PR — documented here for visibility.

### DIFF
--- a/scripts/detect-tools.sh
+++ b/scripts/detect-tools.sh
@@ -56,11 +56,11 @@ else
         while [[ "$1" == -* ]]; do _jq_flags="$_jq_flags $1"; shift; done
         local _jq_query="$1"
         local _jq_file="$2"
-        $PYTHON -c "
+        $PYTHON - "$_jq_file" "$_jq_query" << 'PYEOF' 2>/dev/null
 import json, sys
 try:
-    data = json.load(open('$_jq_file'))
-    keys = '$_jq_query'.strip('.').split('.')
+    data = json.load(open(sys.argv[1]))
+    keys = sys.argv[2].strip('.').split('.')
     val = data
     for k in keys:
         if k and isinstance(val, dict):
@@ -72,7 +72,7 @@ try:
     print(val if isinstance(val, (str, int, float, bool)) else json.dumps(val))
 except Exception:
     sys.exit(0)
-" 2>/dev/null
+PYEOF
     }
     JQ="_jq_fallback"
 fi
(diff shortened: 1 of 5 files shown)


fdaviddpt:
3 changes, 9 new tests, 367 existing tests pass. Reviewer found no blockers: safe_eval RCE eliminated via printf-v indirect assignment, trap path correctly single-quoted for TMPDIR-with-spaces, _jq_fallback path injection closed via sys.argv. One pre-existing dry-run test failure (no session .jsonl in bare checkout) is unrelated to this PR.

Issue #67, opened 2026-05-24
Security: tighten trust boundary on config-driven exec and git push


fdaviddpt:
Internal security audit, grouped findings on "anyone who can write to config/hooks gets code exec."

### 2.1 Git backup pushes to attacker-controlled remote without validation — **High**

`hooks.d/after_save/50-git-backup.sh:106`: `git push` runs with no check on the remote URL.

**Scenario:** attacker writes to `~/.remember/config.json` (via a malicious tool called by Claude, compromised project, or symlink). Sets `data_dir` to a path whose parent is a git repo with `origin = https://attacker.example/exfil.git`. Every session save silently pushes all memory — identity, conversation summaries, project paths, anything discussed — to the attacker.

`GIT_TERMINAL_PROMPT=0` prevents credential prompts but doesn't validate the remote.

**Fix:**
- On first push, log the remote URL prominently. On subsequent pushes, abort if the URL changed without explicit user confirmation.
- Or: validate against an allowlist pattern (`file://`, `ssh://github.com/*`, `https://github.com/*`) configurable via `git_backup.remote_allowlist`.

### 2.2 `dispatch()` executes anything dropped in `hooks.d/` — **Medium**

`scripts/log.sh:163`: iterates `hooks.d/<event>/*` and runs any file with `-x`. No ownership check, no signature verification.

Plugin cache (`~/.claude/plugins/cache/*/`) is user-writable by design. Any process that can write there — malicious npm package, compromised Claude tool, browser extension with FS access — gets arbitrary code execution on every session save and start.

**Fix:** before executing a hook, verify ownership and that it's not world-writable:

```sh
[ "$(stat -f %u "$hook" 2>/dev/null || stat -c %u "$hook")" = "$(id -u)" ] || continue
[ -z "$(find "$hook" -perm -002)" ] || continue
```

Cheap. Eliminates the dropped-executable attack.

### 2.3 Document the trust model

README should state explicitly: "Anyone who can write to `~/.remember/config.json`, the plugin's `hooks.d/`, or the configured git remote has code execution / data exfil with the user's privileges."

Users installing a memory plugin don't intuit that — making it explicit lets them reason about which directories to lock down.

---

2.1 is the most impactful — silent exfil channel via a config file edit. 2.2 closes the hooks.d hole. 2.3 is documentation only.

PR #71, opened 2026-05-24
security: tighten trust boundary on git push and hooks.d exec (#67)

fdaviddpt:
## What this closes

Three security findings from the internal audit in #67:

**2.1 — Silent exfil via poisoned remote (High):** An attacker who writes to `~/.remember/config.json` could redirect `data_dir` to a repo with `origin = attacker-controlled`. Every session save would silently push all memory there. Fixed by recording the remote URL on first push (`~/.remember/.git-backup-remote`) and aborting if it changes. Set `git_backup.allow_remote_change = true` to re-point intentionally.

**2.2 — Dropped executable in hooks.d (Medium):** `dispatch()` ran any world-writable or foreign-owned file. Any process that can write to `~/.claude/plugins/cache/` got code execution on every save. Fixed by checking ownership (`stat -f %u` / `stat -c %u`, BSD/GNU portable) and the world-writable bit before executing each hook. Mismatched hooks are skipped with a logged warning.

**2.3 — Undocumented trust model:** README now has a `## Trust Model` section (after Install, before Changelog) stating plainly what write access to config/hooks/remote means.

## Backward compatibility

- Users with stable remotes and owned hooks: no behavior change.
- Users intentionally re-pointing their backup remote: set `git_backup.allow_remote_change = true` once, then set it back.
- The `.git-backup-remote` state file is written on the first push after upgrade — existing users see one extra log line: "git backup configured to push to: <url>".

## Tests added

- `TestGitBackupRemoteValidation` (4 tests): first push records URL; second push to same URL succeeds; push to changed URL aborts; `allow_remote_change=true` override works.
- `TestDispatchOwnershipChecks` (3 tests): owned+non-world-writable executes; world-writable skipped; unowned skipped.
- Total: 365 passed (up from 327).

## Trade-offs

- Ownership check uses a fake-stat approach in tests (can't `chown` without root); the production path is straightforward `stat` output comparison.
- World-writable check uses `find -maxdepth 0 -perm -002` — portable across BSD and GNU, avoids `stat` format inconsistencies for permission bits.
- The dry-run save failure (test 7 in `run-tests.sh`) is pre-existing on `main` and unrelated to this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -326,6 +326,7 @@ git remote add origin git@github.com:youruser/remember-backup.git  # private rep
 cat > .gitignore <<'EOF'
 .git-backup.lock
 .last-git-backup-ts
+.git-backup-remote
 */logs/
 */tmp/
 EOF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,19 @@ Look at the `version` field in `.claude-plugin/plugin.json`. The plugin location
 
 **The story behind it:** [I built a memory system I'll never remember building](https://max.dp.tools/posts/134-i-built-a-memory-system-ill-never-remember-building.php) — by Max, the AI that designed it and doesn't remember.
 
+## Trust Model
+
+This plugin runs with your full shell privileges. Before installing or upgrading, understand what you're trusting:
+
+- **`~/.remember/config.json`** — Any process that can write this file can redirect the git backup remote to an attacker-controlled URL, causing all memory to be silently exfiltrated on every session save.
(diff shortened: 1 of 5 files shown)


fdaviddpt:
3-line summary: Closes the silent exfil channel (poisoned config.json → attacker remote) by validating the push target URL on every backup and aborting if it changed. Locks down the dropped-hook exec vector by checking ownership and world-writable bit in dispatch() before running any hook. Post-review fix: `.git-backup-remote` is now gitignored so the state file itself isn't committed to the backup repo. 365 tests pass (+38 new).

Issue #68, opened 2026-05-24
Security: set `umask 077` at pipeline entry, files created world-readable today


fdaviddpt:
Internal security audit finding.

**Severity: Low**

`scripts/log.sh:102`, `scripts/bootstrap-dirs.sh:74`: log files and memory directories are created with `>>` and `mkdir -p`, both of which inherit the umask. On systems with a permissive umask (e.g. `022`, the default on most distros), files are created `-rw-r--r--` — world-readable.

Contents that leak via permissive permissions:
- `memory-YYYY-MM-DD.log`: project paths, branch names, token counts, timing data — enough to fingerprint user activity
- `.remember/` memory files: identity, conversation summaries, project context
- Temp files in `$TMPDIR`: extraction payloads, config merges

On a multi-user machine (shared dev box, CI runner, jumphost), any other user can read all of this.

**Fix:** add one line at the top of `scripts/resolve-paths.sh` (sourced by every entry point):

```sh
umask 077
```

Protects all files created downstream — logs, temp files, memory files — without per-file changes.

Alternative if `umask 077` breaks anything: use `install -m 600 /dev/null "$logfile"` for explicit creation, but the global `umask` is simpler and catches future code that creates files without thinking about modes.

PR #72, opened 2026-05-24
security: set umask 077 to prevent world-readable files (#68)

fdaviddpt:
## Summary

- Adds `umask 077` at the top of `scripts/resolve-paths.sh`, which is sourced by every pipeline entry point (`post-tool-hook.sh`, `session-start-hook.sh`, `user-prompt-hook.sh`, `save-session.sh`, `run-consolidation.sh`)
- Protects all downstream-created files: daily log files (`memory-YYYY-MM-DD.log`), hook error logs, `.remember/` directories, lock/marker files in `.remember/tmp/`, and `mktemp` temp files in `$TMPDIR`
- Adds 6 tests in `tests/test_umask.py` verifying umask is 077 after source, files are mode 600, directories are mode 700, log files are 600, and temp files are 600

## Scope

Every file and directory created after `resolve-paths.sh` is sourced inherits the restrictive umask. That covers the full pipeline surface: logs that contain project paths/branch names/token counts, `.remember/` memory files with identity and conversation summaries, and temp extraction/prompt payloads in `$TMPDIR`.

## Entry point coverage

All 5 entry points that source `resolve-paths.sh` are covered by the single change:
- `post-tool-hook.sh` — line 31
- `session-start-hook.sh` — line 40
- `user-prompt-hook.sh` — line 31
- `save-session.sh` — line 57
- `run-consolidation.sh` — line 38

No entry point needed a direct `umask` addition.

## Backward compatibility

`umask 077` is strictly more restrictive than the status quo — it never grants permissions, only withholds them. Existing files are unaffected. The only observable change: newly created files are owner-only by default. No application logic reads file permissions, so no code paths break. CI environments running as a single user are unaffected.

## Edge cases considered

- **CI runners**: Single-user containers — umask 077 has no effect on functionality, slightly tightens sandbox
- **Multi-user dev boxes / jumphosts**: This is exactly the threat model in issue #68 — fixed
- **Windows (Git Bash / MSYS / Cygwin)**: `umask` is a no-op on Windows file systems; the existing Windows normalization path is unaffected
- **Pre-existing dry-run test failure**: `save-session.sh --dry: exit 0` was already failing on `main` before this PR (confirmed by running the suite on the unmodified branch). Not introduced here.

## Trade-offs

The alternative suggested in the issue — `install -m 600 /dev/null "$logfile"` per-file explicit creation — would require touching every file-creation site and would miss future code that creates files without thinking about modes. The global `umask` is the right lever: one line, total coverage, self-maintaining.

### DIFF
--- a/scripts/resolve-paths.sh
+++ b/scripts/resolve-paths.sh
@@ -30,6 +30,14 @@
 #
 # ============================================================================
 
+# --- Restrict file creation permissions ---
+# Prevent log files, memory files, and temp files from being world/group readable.
+# On multi-user machines (shared dev box, CI runner, jumphost) the default umask
+# (022) creates files as -rw-r--r--, leaking project paths, branch names, token
+# counts, and memory contents to any local user.  Setting 077 here covers every
+# downstream file created after this source: logs, .remember/ dirs, TMPDIR temps.
+umask 077
+
 # --- Resolve PIPELINE_DIR (where the plugin code lives) ---
 #
 # Priority:
--- a/tests/test_umask.py
+++ b/tests/test_umask.py
@@ -1,0 +1,182 @@
+"""Tests for umask 077 enforcement in the pipeline entry point.
+
+Issue #68: files created with the default umask (022) are world-readable on
+multi-user machines. resolve-paths.sh now sets umask 077 so that all downstream
+files (logs, memory dirs, temp files) are created with mode 600/700.
+"""
+
+import os
+import stat
+import subprocess
+import tempfile
+from pathlib import Path
+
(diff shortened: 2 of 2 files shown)


fdaviddpt:
**Fix summary:** Added `umask 077` to `scripts/resolve-paths.sh` (sourced by all 5 entry points). All downstream files — logs, memory dirs, temp files — now created mode 600/700 instead of 644/755. Six new tests in `tests/test_umask.py` verify exact modes. Pre-existing dry-run test failure on `main` is unrelated to this change.

Issue #69, opened 2026-05-24
Security: data-integrity nits in consolidation and git-backup locking


fdaviddpt:
Internal security audit, two low-impact integrity issues grouped.

### 4.1 STAGING path unescaping is asymmetric — **Medium**

`scripts/run-consolidation.sh:89`: staging paths from Python are extracted with

```sh
sed "s/^STAGING=//;s/^'//;s/'$//"
```

This undoes `_shell_escape()`'s outer single-quote wrapping but doesn't decode the `'\''` idiom used for embedded single quotes. A staging filename containing a single quote produces a corrupt path — `mv` then points at the wrong file (or fails), leaving the memory file un-renamed to `.done.md` and re-processed indefinitely.

Not directly exploitable for code execution, but a data-loss path on filesystems where single quotes in filenames are valid.

**Fix:** write staging paths to a temp file (one per line, NUL-separated), and read with `while IFS= read -r -d '' path; do ...`. Or move to JSON IPC alongside the broader `safe_eval` refactor.

### 4.2 Stale lock TOCTOU in git-backup — **Low**

`hooks.d/after_save/50-git-backup.sh:61-62`:

```sh
rm -f "$LOCK_FILE"
( set -o noclobber; echo $$ > "$LOCK_FILE" ) 2>/dev/null || exit 0
```

Between `rm` and the `noclobber` re-acquire, another concurrent session can win the lock. Both then proceed.

Benign in practice — worst case is two concurrent commits to disjoint slug subtrees (which git serializes via its own index lock). Worth a code comment acknowledging the race + why it's acceptable.

**Fix:** add a comment, or use `flock(1)` on a separate lock file (no TOCTOU — `flock` is atomic). The latter is the structurally correct fix and removes the race.

---

Both are low-impact today. 4.1 is the one to actually fix; 4.2 is acknowledge-or-flock.

PR #73, opened 2026-05-24
security: harden STAGING parsing and git-backup lock (#69)

fdaviddpt:
## Summary

Fixes both issues from #69.

**4.1 — STAGING path asymmetric unescaping (Medium)**

The previous shell-side parsing used `sed "s/^STAGING=//;s/^'//;s/'$//"` to decode staging paths emitted by Python. This stripped the outer single-quote wrapping but did not decode `'\''` sequences, so any staging filename containing a single quote would produce a corrupt path — `mv` would fail and the file would be re-processed on every future consolidation run (data-loss loop).

Fix: Python's `cmd_consolidate` now writes all staging paths to a NUL-separated temp file and emits its path as `STAGING_PATHS_FILE=`. The shell rename loop reads with `while IFS= read -r -d '' path; do ...`, which handles single quotes, spaces, and all shell metacharacters without any encoding layer.

**4.2 — Stale lock TOCTOU in git-backup (Low)**

The original pattern (`rm -f` + `noclobber` re-acquire) has a window between the `rm` and the re-acquire where two concurrent sessions can both proceed. Worst case is two commits to disjoint slug subtrees (git serializes these via its own index lock — benign in practice).

Fix: detect `flock(1)` at runtime. If available, use `exec 9>"$LOCK_FILE" && flock -n 9` — atomic, no race. If unavailable (macOS without util-linux, rare), fall back to the noclobber pattern with an inline comment acknowledging the benign race.

## Test plan

- `test_cmd_consolidate_staging_paths_file_handles_special_chars` — verifies NUL-separated paths file round-trips filenames with spaces correctly
- `test_cmd_consolidate_with_staging_files` — updated: asserts `STAGING_PATHS_FILE=` is present and no `STAGING=` lines remain
- `test_flock_concurrent_only_one_wins` — launches two hook invocations concurrently and asserts exactly one commit results (skipped on macOS without flock)
- Existing `test_lock_contention_skips` and `test_stale_lock_takeover` skip when flock is present (flock supersedes that code path)
- All 42 tests pass (1 skipped — flock test on macOS); pre-existing dry-run and coverage failures unchanged

### DIFF
--- a/hooks.d/after_save/50-git-backup.sh
+++ b/hooks.d/after_save/50-git-backup.sh
@@ -48,18 +48,36 @@ if [ -f "$COOLDOWN_MARKER" ]; then
     fi
 fi
 
-# ── Lock (atomic via noclobber) ───────────────────────────────────────────────
+# ── Lock ─────────────────────────────────────────────────────────────────────
+# Prefer flock(1): it acquires atomically on an open fd, eliminating the
+# TOCTOU window between rm and noclobber re-acquire.  On macOS without
+# util-linux flock, we fall back to the noclobber pattern; the race is benign
+# (worst case: two concurrent commits to disjoint slug subtrees, which git
+# serializes via its own index lock).
 LOCK_FILE="$REPO_ROOT/.git-backup.lock"
-if ! ( set -o noclobber; echo $$ > "$LOCK_FILE" ) 2>/dev/null; then
-    LOCK_PID=$(cat "$LOCK_FILE" 2>/dev/null)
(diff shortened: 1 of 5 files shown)


fdaviddpt:
Fix summary: STAGING parsing replaced with NUL-separated temp file (eliminates single-quote data-loss loop); git-backup lock uses flock when available (removes TOCTOU race) with noclobber fallback for macOS. 42 tests pass, 1 new test added per fix.

PR #88, opened 2026-05-30
feat(save-session): honor $REMEMBER_BRANCH env var when $PROJECT_DIR is not a git repo

gandfrey:
## Summary

When Claude Code runs from a directory that isn't a git repo (e.g. `\$HOME`, a scratch dir, anywhere the user hasn't `cd`'d into a project first), the `## HH:MM | <branch>` identity slot in every `today-*.md` entry falls back to the literal string `unknown`. The slot collapses across all entries from that working directory — and across multiple Claude Code instances that each run from non-git CWDs, the daily log becomes indistinguishable per writer.

This PR adds `\$REMEMBER_BRANCH` as the highest-priority source for that slot. Backward compatible — zero behavior change when the env var is unset.

```diff
- BRANCH=\$(cd \"\$PROJECT_DIR\" && git branch --show-current 2>/dev/null || echo \"unknown\")
+ BRANCH=\"\${REMEMBER_BRANCH:-\$(cd \"\$PROJECT_DIR\" && git branch --show-current 2>/dev/null || echo \"unknown\")}\"
```

**Use case I hit:** a multi-instance memory protocol where `today-*.md` entries from two Claude Code instances (one on a laptop, one on a Lightsail box) get rsync'd into a shared `_inbox/`. Both ran Claude Code from `\$HOME` — neither was a git repo — so every entry's identity slot read `unknown`. With `export REMEMBER_BRANCH=laptop` / `export REMEMBER_BRANCH=cloud` in each shell rc, the slot becomes a who-wrote-this tag and the log becomes useful again.

## Changes

- `scripts/save-session.sh` — one-line change to the `BRANCH=` line (line 150) using `\${REMEMBER_BRANCH:-…}`. The `:-` form (not bare `-`) intentionally treats `REMEMBER_BRANCH=` as unset, so an accidental empty export in a shell rc doesn't propagate an empty header — it walks the existing git/unknown fallback chain.
- `tests/test_save_session_branch_override.py` — new test file pinning the four-case truth table (env-set + git-repo, env-unset + git-repo, env-unset + no-git, env-set-empty). Snapshots the live `BRANCH=` line out of `save-session.sh` rather than reasserting a copy, so future edits to the line fail this test on intent rather than on string-equality drift. Includes a guard that the line uses the safe `:-` operator, so a regression back to bare `-` (which would propagate the empty string) surfaces in code review even before tests run.
- `README.md` — new \"Environment variables\" subsection under Configuration documenting `REMEMBER_BRANCH`, `REMEMBER_DEBUG`, and `REMEMBER_TZ`.
- `CHANGELOG.md` — Unreleased entry.

## Test plan

- [x] `pytest` passes locally — **396 passed, 1 skipped** (was 391 + 5 new in this PR).
- [x] New behavior covered by a test — four-case truth table + the `:-`-operator guard, all in `tests/test_save_session_branch_override.py`.
- [x] Touched a hook (`scripts/save-session.sh`) — manually verified by applying the same one-line patch to the locally-cached plugin (`~/.claude/plugins/cache/claude-plugins-official/remember/0.7.3/scripts/save-session.sh`) and exporting `REMEMBER_BRANCH=laptop` in `~/.zshrc`. The local change is what motivated submitting this upstream so it survives plugin updates.
- [ ] Path resolution untouched — N/A.

**One transparency note:** I have not yet round-tripped a full `Stop → save-session → today-*.md write` against this exact branch after running `claude` in a fresh shell that picked up the new env var (the env var was set after this session started). The behavior is exercised by the new tests; the live round-trip is whatever the next session start does, which I'd verify post-merge.

## Checklist

- [x] Docs updated — env-var subsection added to `README.md` Configuration block; no new config key, hook, or skill was added (just a new env-var input to an existing hook).
- [ ] Version bumped in `.claude-plugin/plugin.json` — left to maintainer at release time.
- [x] Changelog entry — added to `CHANGELOG.md` (existing PRs touch CHANGELOG.md; the template wording mentions `README.md` but I followed the actual project convention).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **`REMEMBER_BRANCH` env var override** — `scripts/save-session.sh` now honors `$REMEMBER_BRANCH` when computing the `## HH:MM | <branch>` identity slot of each daily-log entry. Falls back to the existing `git branch --show-current` lookup, then the literal `"unknown"` if no git repo is present. Use case: running Claude Code from `$HOME` (or any non-git directory) collapses the identity slot to `unknown` on every entry, which makes log entries indistinguishable across instances. Export `REMEMBER_BRANCH=laptop` / `cloud` / `staging` / `$HOSTNAME` in your shell rc and the slot becomes a useful per-instance tag. Documented in `README.md` Configuration → Environment variables.
+
+### Tests
+
+- New `tests/test_save_session_branch_override.py` — pins the four-case truth table for the `BRANCH=` line in `save-session.sh`: env-set + git-repo (env wins), env-unset + git-repo (git wins), env-unset + no-git (`unknown` fallback), env-set-to-empty + no-git (`:-` treats empty as unset, falls back to `unknown`). Snapshots the line out of the live `save-session.sh` rather than re-asserting a copy, so the test fails loudly if the line is ever edited without updating the test.
+
 ## [0.7.3] — Windows save pipeline shell↔Python bridge
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


fdaviddpt:
Thanks @gandfrey — clean fix. The `:-` choice (treating empty `REMEMBER_BRANCH=` as unset so an accidental export doesn't propagate an empty header) is exactly right, and snapshotting the live `BRANCH=` line out of `save-session.sh` so edits fail on intent rather than string drift is a sharp test pattern.

One note for a follow-up if multi-instance identity grows legs: the `<branch>` slot is being repurposed as a who-wrote-this tag, which works but is semantically a branch field — worth its own identity field down the line rather than overloading this one. Not blocking. Merging now.

Issue #87, opened 2026-05-30
perf(haiku): isolate the headless `claude -p` summarization subprocess


fdaviddpt:
## Context

`pipeline/haiku.py::call_haiku()` spawns a headless `claude -p` subprocess per summarization. It's a tool-less, one-shot, `--max-turns 1`, JSON-output call — it only needs to summarize the prompt and return text. Current command:

```
claude -p PROMPT --output-format json --model haiku --max-turns 1 --allowedTools ""
```

It runs with `cwd=tempfile.gettempdir()` and `CLAUDECODE` stripped from the env.

The subprocess is doing far more than the job requires: it loads the user's global `~/.claude/CLAUDE.md`, settings, hooks, and MCP servers into a call that never needed any of them. This is wasted input tokens, wasted startup, **output contamination** (global instructions — e.g. custom output styles — leak into summaries), and a real eviction bug (see #86 / below).

All flags below verified against `claude --help` on **Claude Code v2.1.156**. Latency figures are intentionally omitted — none have been benchmarked. The point of this issue is to land the safe wins and benchmark the rest.

## The original trigger — PR #86 (withdrawn)

PR #86 added `--strict-mcp-config` to fix a concrete bug: without it, the subprocess boots the user's globally-configured MCP servers. For single-consumer channel plugins (Telegram = one `getUpdates` slot per bot) the spawned poller **evicts the user's live session poller**, then exits and clears the slot → the user's channel goes silent until a manual `/mcp` reconnect. Diagnosed on a 24/7 session where every Telegram death lined up with a `[haiku] consolidation` event. PR was withdrawn; the fix never landed. It should — either via `--strict-mcp-config` or `--bare` (below).

## Tier 1 — pure wins (no auth change, ship these)

| Flag | Effect | Notes |
| --- | --- | --- |
| `--strict-mcp-config` | With no `--mcp-config` given, loads **zero** MCP servers. | Fixes the #86 eviction bug. No caller passes an MCP tool (only real caller `consolidate.py` passes no tools; tests use `Read`/`Write`). Safe. |
| `--setting-sources project` (or narrowest valid) | Restricts which settings JSON files load (`user`/`project`/`local`). | **Caveat:** does NOT block CLAUDE.md discovery or hooks — only settings files. Cannot be set to empty. Partial mitigation only; `--bare` is the real fix (Tier 3). |
[...]

PR #93, opened 2026-06-05
feat(haiku): isolate one-shot subprocess (Tier-1 from #87)

sergeclaesen:
Implements the highest-confidence **Tier-1 "pure wins"** from #87 — isolates the one-shot summarization/consolidation subprocess so it stops doing work it never needs.

## What changed

Adds two `--print`-compatible, no-auth-change flags to all three `claude -p` call sites (`scripts/save-session.sh` haiku + ndc, and `pipeline/haiku.py`):

| Flag | Why |
| --- | --- |
| `--no-session-persistence` | The summarization call is one-shot and never resumed, but it currently writes a session record to disk on every run. On a 24/7 setup those records sync into the desktop app's **Recents** list and flood it — hundreds per day on busy interactive days, all landing in `-private-tmp` / `-private-var-folders-…-T` project dirs (the `cd /tmp`). Pure win. |
| `--exclude-dynamic-system-prompt-sections` | Moves the dynamic preamble (cwd, env, git status, OS) out of the static system prompt — input the summarization call doesn't need. |

I deliberately left the other Tier-1 items out to keep this tight and unambiguous:
- `--strict-mcp-config` already landed in `save-session.sh`.
- `--setting-sources project` is caveated as partial in #87 and `project` is the wrong scope for a `/tmp` cwd with no project — needs a decision.
- `--max-budget-usd` needs a value choice — maintainer call.

## Verification

- Both flags verified against `claude --help` on **Claude Code v2.1.163**.
- Smoke-tested the exact production invocation (`--model haiku --max-turns 1 --output-format json --no-session-persistence --exclude-dynamic-system-prompt-sections --mcp-config '{}' --strict-mcp-config`): JSON output intact (`is_error: false`), and **zero** new session records written to `~/.claude/projects/`.
- Added test assertions for both flags in `tests/test_haiku.py`.
- Full suite green: `396 passed, 1 skipped`. `bash -n` clean on `save-session.sh`.

Refs #87.

### DIFF
--- a/pipeline/haiku.py
+++ b/pipeline/haiku.py
@@ -57,6 +57,8 @@ def call_haiku(
         "claude",
         "-p", prompt,
         "--output-format", "json",
+        "--no-session-persistence",
+        "--exclude-dynamic-system-prompt-sections",
         "--model", "haiku",
         "--max-turns", "1",
         "--allowedTools", ",".join(tools) if tools else "",
--- a/scripts/save-session.sh
+++ b/scripts/save-session.sh
@@ -175,6 +175,7 @@ CLEANUP_FILES+=("$HAIKU_STDERR")
 HAIKU_JSON=$(cd /tmp && env -u CLAUDECODE claude -p \
     --model haiku --allowedTools "" --max-turns 1 \
     --output-format json \
+    --no-session-persistence --exclude-dynamic-system-prompt-sections \
     --mcp-config '{"mcpServers":{}}' --strict-mcp-config \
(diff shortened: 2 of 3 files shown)


fdaviddpt:
Merged — thank you! 🙏 Isolating the summarizer subprocess (`--no-session-persistence` + `--exclude-dynamic-system-prompt-sections`) stops the session-record flood at the source, and your field report with the per-day numbers made the case airtight. Genuinely appreciated.

PR #99, opened 2026-06-17
fix: resolve empty timezone to system-local, never bare TZ="" (UTC)

kristian-presso:
## Problem

The date-resolution sites in `save-session.sh`, `session-start-hook.sh`, and
`user-prompt-hook.sh` use the unguarded pattern:

```sh
TZ="$REMEMBER_TZ" date +%Y-%m-%d
```

When `REMEMBER_TZ` is empty - the default when no `timezone` is set in config, which is
the common case - this expands to `TZ="" date`. On macOS/BSD an empty `TZ` resolves to
**UTC**, not system-local. So an unconfigured user west of UTC gets daily/rotating files
that roll at the wrong local midnight: a write just after local midnight lands in the
previous day's file. It is silent and only visible during the local-to-UTC offset
window, so it survives casual testing.

`log.sh` already solved this correctly with the `_remember_date` helper:

```sh
_remember_date() {
  if [ -n "$REMEMBER_TZ" ]; then TZ="$REMEMBER_TZ" date "$@"; else date "$@"; fi
}
```

...but three scripts that `source log.sh` still call the raw pattern instead of the
helper.

## Fix

Re-point every unguarded `TZ="$REMEMBER_TZ" date` call at the existing `_remember_date`
helper (already in scope in each of these files via `source log.sh`). Empty timezone now
falls back to system-local; an explicit `timezone` override still works unchanged. No
behaviour change for users who have set a timezone.

- `scripts/save-session.sh` (3 sites)
- `scripts/session-start-hook.sh` (1 site)
- `scripts/user-prompt-hook.sh` (2 sites)

`log.sh`'s own helper body is intentionally untouched (its `TZ="$REMEMBER_TZ" date` is
the correct non-empty branch).

## Verify

With no `timezone` configured, run an affected path during the local-to-UTC offset
window and confirm it prints the host-local date, not UTC.

### DIFF
--- a/scripts/save-session.sh
+++ b/scripts/save-session.sh
@@ -64,7 +64,7 @@ LOCK_FILE="${REMEMBER_DIR}/tmp/save.lock"
 MEMORY_FILE="${REMEMBER_DIR}/now.md"
 LAST_SAVE_FILE="${REMEMBER_DIR}/tmp/last-save.json"
 COOLDOWN_MARKER="${REMEMBER_DIR}/tmp/last-save-ts"
-TODAY_DATE=$(TZ="$REMEMBER_TZ" date +%Y-%m-%d)
+TODAY_DATE=$(_remember_date +%Y-%m-%d)
 CLEANUP_FILES=()
 
 # Remove lock file and all accumulated temp files on exit.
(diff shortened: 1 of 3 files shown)


fdaviddpt:
Merged — thank you! 🙏 The empty-`TZ`→system-local fix was a real correctness bug for anyone west of UTC, and routing it through the existing `_remember_date` helper was exactly the right call. Clean, well-scoped PR — exactly the kind we love to get.

PR #110, opened 2026-06-22
fix: stop consolidation/NDC from writing slop into memory files

slcc2c:
## What

Three defects let the Haiku consolidation/NDC pipeline write machine-noise into the long-lived memory files (`recent.md`, daily staging). All three were observed producing real artifacts in a live `~/.remember/` tree.

### 1. NDC compression appends a self-narration trailer
`prompts/compress-ndc.prompt.txt` ended with *"No preamble. Just the compressed output."* — it bans a **preamble** but not a **trailer**, so Haiku obeys literally and appends a `**Compression applied:**` block narrating what it shortened. Fixed by banning trailers/meta-commentary explicitly.

### 2. Consolidation leaks a code fence + a duplicate `# Recent`
- `prompts/consolidate-staging.prompt.txt` showed the required output envelope (and the Step-3 identity-candidate example) **wrapped in ``` code fences**. The model imitates the fence and emits a stray ``` as the first line of its output. Now shown as raw / indented text, with an explicit "no code fences, nothing before/after the delimiters" instruction.
- `pipeline/consolidate.py::parse_consolidation_response` did a naive `recent.startswith("# Recent")` check. With a leading stray fence, the check fails and a **second** `# Recent` is prepended → duplicate heading + orphan fence in `recent.md`. Now strips leading/trailing fence lines first (`_strip_code_fences`).
- Replaced the rule *"Preserve the `# Recent` and `# Archive` headers"* with *"output exactly one of each"* — the template and the parser already supply them; the old wording nudged a duplicate header into the body.

## Tests
Adds regression coverage for fence-wrapped envelopes (`test_parse_strips_wrapping_code_fences`, `test_strip_code_fences_helper`). Full suite: **450 passed, 41 skipped, 98% coverage** (was 448).

## Deliberately *not* changed (FYI)
The `else` fallback in `parse_consolidation_response`, combined with `_is_valid_consolidation` accepting entry-header-only output, can in principle write a delimiter-less response straight to `recent.md`. I left it untouched because your tests at `tests/test_consolidate.py:177-180` and `:135-143` assert this salvage behavior is intentional. Flagging in case the trade-off is worth revisiting — but out of scope here.

### DIFF
--- a/pipeline/consolidate.py
+++ b/pipeline/consolidate.py
@@ -32,6 +32,24 @@ from .types import ConsolidationResult, TokenUsage
 _ENTRY_HEADER = re.compile(r"(?m)^## (\d{2}:\d{2}|Week of |\d{4}-\d{2}-\d{2})")
 
 
+def _strip_code_fences(text: str) -> str:
+    """Strip leading/trailing Markdown code-fence lines from a parsed section.
+
+    Haiku sometimes wraps its whole response in a ``` ... ``` block. Left in,
+    a leading fence pushes the real ``# Recent``/``# Archive`` header off the
+    first line, defeating the ``startswith`` check in
+    :func:`parse_consolidation_response` — so a second header gets prepended,
+    producing an orphan-fence + duplicate-heading artifact in the saved file.
+    Removing fence-only lines from both ends makes the header check robust.
+    """
+    lines = text.splitlines()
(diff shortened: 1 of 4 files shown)


slcc2c:
Closing — opened in error without proper sign-off on my end. Apologies for the noise.

Issue #96, opened 2026-06-11
save-session: no cap on extract size → oversized Haiku prompt exceeds context window → silent rotation stop (complements #94)


selvi5006-commits:
## Summary

A single long-lived session can grow an extract far larger than Haiku's context window. `save-session.sh` embeds the **entire** extract into the Step-4 prompt with no size cap, so the `claude -p --model haiku` call fails (input exceeds context length), `set -e` aborts the save, and the daily rotation (`now.md` → `today-YYYY-MM-DD.md`) silently stops. It is **self-reinforcing**: a failed save never advances the saved position, so the same session re-extracts the full transcript on every subsequent save and fails identically — until an unrelated small session happens to run.

Confirmed on marketplace `remember@claude-plugins-official` **v0.7.3**, Windows / Git Bash.

This is **related to but distinct from #94**. #94 attributes `FAILED at line ~179` to machine-load-induced *hangs/timeouts* plus the dead `set -e` error handler. My incident is not a hang — every failing save logs an oversized extract and fails in ~25s, not a 180s timeout. The two are complementary: a **size cap prevents the failure**; #94's `|| HAIKU_EXIT=$?` + `timeout` change makes any *residual* failure visible and non-blocking. Neither alone is sufficient.

---

## Evidence (from `.remember/logs/memory-*.log`)

Daily rotation went dark for two full days. Every save against the offending session logged:

```
[extract] 5389 exchanges (292 human)
[haiku] calling (branch: ...)
[rotate] ERROR: tar failed for 30 logs        # non-fatal red herring (log.sh:220 else-branch returns 0)
[error] FAILED at line 175 (exit 1)
```

Meanwhile *small* sessions on the same machine saved fine, interleaved with the failures:

```
20:34  [extract] 33 exchanges (2 human)   → OK
20:57  [extract] 5389 exchanges (292 human) → FAILED at line 175
20:59  [extract] 51 exchanges (5 human)   → OK, [write] appended
```

5389 exchanges is well past Haiku's 200K-token window, so the CLI rejects the input and exits non-zero. Because the failing line is the `HAIKU_JSON=$(... < "$TMP_PROMPT")` command substitution, `set -e` trips there and the generic `ERR` trap logs `FAILED at line 175` — the helpful `if [ "$HAIKU_EXIT" -ne 0 ]` diagnostic never runs (the dead-handler issue, already noted in #94), which is why the real "context length exceeded" error is invisible in the logs.

---

## Root cause

[...]

PR #123, opened 2026-06-29
Bound consolidation prompt size to prevent context-window overflow (#122)

### LINKED ISSUE #122: Consolidation prompt is uncapped — overflows the model window and stalls saves (twin of #96)
## Summary

v0.8.2 (#96) added an oversized-extract guard to the **save** path (`extract_max_bytes` → `cmd_build_prompt`), but the **consolidation** path has the same unbounded-input flaw and was never capped. `consolidate()` inlines the full staging set + `recent.md` + `archive.md` into one prompt and calls Haiku with no size check. When the assembled prompt exceeds the model window, the nested `claude -p` fails with `Prompt is too long`; `run-consolidation.sh` logs `ERROR` and exits 1, so daily rotation halts — and it is self-reinforcing, because the staging files are not retired and re-feed identically on the next run (exactly the #96 failure mode, one path over).

Observed in the wild: scanning one machine's `~/.claude/projects/`, **57 crashed `sdk-cli` sessions** carried the consolidation system prompt ("You are a memory consolidation agent…") and ended on `Prompt is too long`. (Affects through v0.8.3.)

## Why a cap, not a truncation

The save-path fix tail-truncates the extract — safe there, because the extract only feeds a one-shot summary. Consolidation **rewrites** `recent.md`/`archive.md` from the model output, so tail-truncating the input would feed the model a partial `archive.md` and the dropped head would be permanently lost on the next write. So the guard should **skip** (reusing the existing `ConsolidationSkipped` semantics: leave staging + memory untouched) rather than truncate. Converting a crash into a clean skip is the minimal safe fix; bounding `archive.md` growth so consolidation can always run is the deeper follow-up (see end).

## Proposed fix

New `thresholds.consolidate_max_bytes` (default `600000`, `0` disables), threaded shell → `cmd_consolidate` → `consolidate()`, mirroring how `extract_max_bytes` is threaded to `cmd_build_prompt`.

**`pipeline/consolidate.py`** — guard before the Haiku call:
```python
def consolidate(staging_contents, recent, archive, max_prompt_bytes=0):
    prompt = build_consolidation_prompt(staging_contents, recent, archive)

    # Oversized-prompt guard. Mirrors extract_max_bytes but SKIPS instead of
    # truncating: consolidation rewrites recent/archive, so truncating the input
    # would permanently drop the dropped head on the next write.
    if max_prompt_bytes > 0:
        prompt_bytes = len(prompt.encode("utf-8"))
        if prompt_bytes > max_prompt_bytes:
            raise ConsolidationSkipped(
                f"consolidation prompt too large ({prompt_bytes} bytes > "
                f"{max_prompt_bytes} cap) -- skipping to avoid a context-window "
                f"overflow; staging + memory left untouched"
            )

    result = call_haiku(prompt, timeout=180)
    ...
```

**`pipeline/shell.py`** — `cmd_consolidate(..., max_prompt_bytes: int = 0)` passes it to `consolidate(...)`; dispatch parses `max_prompt_bytes=int(sys.argv[5]) if len(sys.argv) > 5 else 0`.

**`scripts/run-consolidation.sh`**:
```sh
CONSOLIDATE_MAX_BYTES=$(config ".thresholds.consolidate_max_bytes" 600000)
RESULT=$(... pipeline.shell consolidate "$STAGING_DIR" "$RECENT_FILE" "$ARCHIVE_FILE" "$CONSOLIDATE_MAX_BYTES" 2>&1) || { ... }
```

**`config.example.json`** — add `thresholds.consolidate_max_bytes: 600000` + a `_comments` entry.

## Tests
- oversized prompt → raises `ConsolidationSkipped`, and `call_haiku` is **never** called;
- under cap → Haiku called once;
- `max_prompt_bytes=0` → prior behavior preserved;
- dispatch test updated + a new case asserting the 6th argv arg sets the cap.

I implemented and verified this locally against v0.8.3: `tests/test_consolidate.py` + `tests/test_shell.py` 61 passed (py3.12); full suite 490 passed / 41 skipped (the one failure is a pre-existing, host-specific BSD-`mktemp` assertion unrelated to this change). End-to-end: `pipeline.shell consolidate` on a 1.6 MB staging file with a 600 KB cap prints `CONSOLIDATION_STATUS=skip`, exit 0, with `REMEMBER_CLAUDE_BIN=/nonexistent/claude` never triggered (skips before any call). Happy to open a PR if useful.

## Deeper follow-up (separate issue)

The skip prevents the crash but means a project whose `archive.md` has grown past the cap stops consolidating. Root cause is unbounded `archive.md` growth (every run re-feeds the whole archive through the model). Options: rotate/cap `archive.md`, or consolidate staging in size-bounded batches so a single run never approaches the window.

presempathy-awb:
Fixes #122.

## Problem

#96 (v0.8.2) capped the **save** path (`extract_max_bytes` → `cmd_build_prompt`), but the **consolidation** path still inlines the full staging set + `recent.md` + `archive.md` into one prompt and calls Haiku with no size check. A large input overflows the model window (`Prompt is too long`); `run-consolidation.sh` logs `ERROR` and exits 1, so daily rotation halts — and it is self-reinforcing, since staging is not retired and re-feeds identically next run (the #96 failure mode, one path over). Observed as 57 crashed `sdk-cli` sessions carrying the consolidation system prompt on one machine.

## Change

**1. Skip-guard (the direct #122 fix).** New `thresholds.consolidate_max_bytes` (default `600000`, `0` disables), threaded `run-consolidation.sh` → `cmd_consolidate` → `consolidate()`, mirroring how `extract_max_bytes` reaches `cmd_build_prompt`. When the assembled prompt would exceed the cap, consolidation **skips** rather than truncates — because consolidation *rewrites* `recent.md`/`archive.md`, so truncating the input would feed a partial archive and permanently drop the dropped head on the next write. `ConsolidationTooLarge` subclasses `ConsolidationSkipped`, so every existing `except ConsolidationSkipped` handler behaves identically.

**2. Archive rotation (the deeper follow-up noted in #122).** A plain skip prevents the crash but means a project whose `archive.md` has grown past the cap stops consolidating. So when `archive.md` is the oversized bulk, `cmd_consolidate` rotates it to a dated sibling (`archive-YYYY-MM-DD.md`, cold storage — no memory lost) and retries once with a fresh archive, so consolidation keeps progressing. If there is nothing to rotate, or the retry still overflows, or the retry's Haiku call errors, the rotation is undone and the original state is left intact.

I'm happy to **split these into two PRs** (cap first, rotation second) if you'd prefer to land the minimal #122 fix on its own — just say the word.

## Tests

- `tests/test_consolidate.py`: oversized prompt raises `ConsolidationTooLarge` (and is-a `ConsolidationSkipped`) with `call_haiku` never called; under-cap proceeds; `max_prompt_bytes=0` preserves prior behavior.
- `tests/test_shell.py`: dispatch passes the new arg; rotation-then-success; nothing-to-rotate → skip; retry-still-too-large → restore + skip; retry-errors → restore + re-raise.

Full suite green locally (py3.12): **494 passed, 41 skipped**. (One pre-existing failure, `test_bsd_mktemp_no_randomization_with_extension`, is a macOS host-`mktemp` assertion unrelated to this change and untouched by it; it passes on Linux CI.) End-to-end: `pipeline.shell consolidate` on a 1.6 MB staging file with a 600 KB cap prints `CONSOLIDATION_STATUS=skip`, exit 0, never spawning `claude`; and an oversized-archive run rotates then restores `archive.md` byte-for-byte when the retry can't reach the model.

Opened as a **draft** for your review.

### DIFF
--- a/config.example.json
+++ b/config.example.json
@@ -8,7 +8,8 @@
   "thresholds": {
     "min_human_messages": 3,
     "delta_lines_trigger": 50,
-    "extract_max_bytes": 300000
+    "extract_max_bytes": 300000,
+    "consolidate_max_bytes": 600000
   },
   "features": {
     "ndc_compression": true,
@@ -20,6 +21,15 @@
   "reject_pattern": "",
 
   "_comments": {
+    "consolidate_max_bytes": [
+      "Skip-guard cap (UTF-8 bytes) on the assembled consolidation prompt",
+      "(staging + recent.md + archive.md). Default 600000. If the prompt would",
+      "exceed this, consolidation SKIPS that run (staging + memory left intact)",
+      "rather than overflowing Haiku's window with 'Prompt is too long' and",
+      "stalling all saves. Unlike extract_max_bytes the input is NOT truncated,",
+      "because consolidation rewrites recent/archive and truncating would drop",
+      "archived memory. Set 0 to disable the guard."
+    ],
     "model": [
       "Model for the summarize/consolidate calls (claude -p). Default 'haiku'.",
       "Set 'sonnet' (or another tier) to improve salience + compression-cap",
--- a/pipeline/consolidate.py
+++ b/pipeline/consolidate.py
@@ -41,6 +41,17 @@ class ConsolidationSkipped(Exception):
     """
 
 
+class ConsolidationTooLarge(ConsolidationSkipped):
+    """Raised when the assembled prompt exceeds ``max_prompt_bytes``.
+
+    Subclass of ConsolidationSkipped so existing ``except ConsolidationSkipped``
+    handlers keep their "leave everything untouched" behavior unchanged. Callers
+    that CAN shrink the input (e.g. rotate archive.md to a dated sibling and
+    retry with a fresh archive) may catch this subclass specifically to recover
+    instead of skipping forever.
+    """
+
+
 def _is_valid_consolidation(text: str) -> bool:
     """Return True only if the model output is real consolidated memory.
 
@@ -69,6 +80,7 @@ def consolidate(
     staging_contents: dict[str, str],
     recent: str,
     archive: str,
+    max_prompt_bytes: int = 0,
 ) -> ConsolidationResult:
     """Run full consolidation: build prompt, call Haiku, parse response.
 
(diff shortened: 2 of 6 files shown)


fdaviddpt:
Thanks for this — clean fix and genuinely thorough tests. The cap solves the #122 crash-loop directly, and the rotation is well-guarded (restore-on-failure paths all covered). Merging both.

One open item, tracked as a follow-up rather than a blocker: rotated `archive-<date>.md` siblings are preserved on disk but the recall surface doesn't glob them yet, so that memory is cold-storage-only until we wire it up. Filed as #124 — would be a nice second contribution if you're up for it.

PR #127, opened 2026-07-14
fix: key memory to the main checkout for git worktree sessions (#56)

### LINKED ISSUE #56: [Question] plugin with worktress
How do the plugin work if claude code uses worktrees? Does it correctly resolves .remember directory always from root workspace, not a worktree path?

jeongph:
## Summary

Fixes #56 — memory built up during a **git worktree** session was thrown away instead of being kept with the repository.

Claude Code sets `CLAUDE_PROJECT_DIR` to the *worktree* path for worktree sessions, and the plugin derives `REMEMBER_DIR` from it. So:

- **Legacy (default) mode** — memory landed in `<worktree>/.remember`, physically inside the worktree. Because the plugin gitignores it with `*`, a plain `git worktree remove` (no `--force`, no warning) deletes every `now.md` / `today-*.md` / `remember.md` from that session, and none of it is migrated to the main checkout.
- **External mode** — the `{slug}` was computed from the worktree path, producing an orphaned `~/.remember/<slug-of-worktree>` subtree the main checkout's sessions never load.

This matches the observation and fix direction in #56 (thanks @KrzysztofKasprowicz for the report and @dewet22 for the diagnosis).

## Fix

`REMEMBER_DIR` resolution now routes through git's **common dir**. When `PROJECT_DIR` is a linked worktree, memory is keyed to the repo's **main checkout**, so it:

- survives `git worktree remove`, and
- is shared across all worktrees of the same repo (one continuous memory, not a throwaway per worktree).

`CLAUDE_PROJECT_DIR` / `PROJECT_DIR` is left **untouched** — session recovery still resolves transcripts under the worktree slug Claude Code computed, exactly as @dewet22 flagged.

### Fail-safe by design

The redirect only fires when a linked worktree is *positively* identified (git's common-dir ≠ git-dir) **and** the computed main checkout is a real work tree. These all fall through to the unchanged `PROJECT_DIR`:

- ordinary (non-worktree) checkouts
- non-git directories
- bare-repo worktrees (parent of the common dir is not a work tree)
- git without `--path-format` (< 2.31)

So non-worktree and non-git users see byte-for-byte identical behavior.

## Changes

- `scripts/lib-memory-dir.sh` — new `_resolve_memory_project_dir()`; `REMEMBER_DIR` is resolved against it and it is exported as `MEMORY_PROJECT_DIR`.
- `scripts/bootstrap-dirs.sh` — the one-shot legacy→external migration and the `.gitignore` write are keyed to `MEMORY_PROJECT_DIR` so they stay consistent with where `REMEMBER_DIR` now resolves.
- `tests/test_worktree_memory.py` — new: legacy + external resolution in a real worktree, regression guards for non-worktree / non-git, the shared-dir invariant, and that session-start emits `=== HANDOFF ===` pointing at the main checkout (the existing `REMEMBER_ROOT != PROJECT_DIR` guard already covers this once `REMEMBER_DIR` is redirected).
- `CHANGELOG.md`, `README.md` — document the behavior (new "Git worktrees" section).

## Handoff interaction (no extra code needed)

In legacy mode the `=== HANDOFF ===` hint is normally suppressed (#118) because the `/remember` skill's fallback `{project}/.remember/remember.md` already matches. Inside a worktree that fallback would be wrong — but session-start's existing `REMEMBER_ROOT != PROJECT_DIR` guard now fires (since `REMEMBER_ROOT` is the main checkout ≠ the worktree `PROJECT_DIR`), so the hint is emitted and the skill writes the handoff to the main checkout. A test locks this in.

## Testing

- Full suite (CI-equivalent, `pytest` with coverage): **500 passed, 43 skipped, 98.24% coverage**.
- New `tests/test_worktree_memory.py` uses real `git worktree` layouts.
- Manual end-to-end: accumulated memory during a worktree session now survives `git worktree remove` and lands in the main checkout's `.remember/`.

> Windows is skipped for the new tests (POSIX worktree layout + slug assertions), consistent with the existing worktree/path tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Worktree sessions built a throwaway memory that vanished on cleanup** ([#56](https://github.com/Digital-Process-Tools/claude-remember/issues/56)) — the plugin derives `REMEMBER_DIR` from `CLAUDE_PROJECT_DIR`, which Claude Code sets to the *worktree* path for worktree sessions. In the default (legacy) layout that put memory at `<worktree>/.remember` — physically inside the worktree, and gitignored with `*`, so a plain `git worktree remove` (no `--force`, no warning) deleted every `now.md` / `today-*.md` / `remember.md` built up during the session, and none of it was ever migrated to the main checkout. In external mode the `{slug}` was computed from the worktree path, producing an orphaned `~/.remember/<slug-of-worktree>` subtree the main checkout's sessions never loaded. `REMEMBER_DIR` resolution now routes through git's *common dir*: when `PROJECT_DIR` is a linked worktree, memory is keyed to the main checkout, so it survives `worktree remove` and is shared across all worktrees of the repo. `PROJECT_DIR` itself is left untouched (session recovery still resolves transcripts under the worktree slug), and non-worktree / non-git projects behave exactly as before. Reported and diagnosed by [@KrzysztofKasprowicz](https://github.com/Digital-Process-Tools/claude-remember/issues/56) and [@dewet22](https://github.com/Digital-Process-Tools/claude-remember/issues/56).
+
 ## [0.8.4] — Bound consolidation prompt size so a huge archive can't stall saves
 
 ### Fixed
--- a/README.md
+++ b/README.md
@@ -315,6 +315,15 @@ Once `~/.remember/` is a git repo, the `after_save` hook commits each project's 
 
 If you don't want automatic commits, leave `~/.remember/` as a plain directory and commit manually as before.
 
+## Git worktrees
+
+Claude Code sets `CLAUDE_PROJECT_DIR` to the *worktree* path for sessions started inside a [git worktree](https://git-scm.com/docs/git-worktree). Memory is deliberately **not** kept in the worktree — it is keyed to the repository's **main checkout** instead, so that:
+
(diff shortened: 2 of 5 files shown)


fdaviddpt:
Thanks for this, @jeongph — really clean fix. 🙏

Reviewed and verified the core logic rather than just reading it:

- Reproduced `_resolve_memory_project_dir()` under `set -euo pipefail` (bash 5 / git 2.46) for non-git, ordinary checkout, and linked worktree — all resolve correctly, and the `{ read; read; }<<EOF` block doesn't trip `set -e` on EOF.
- Non-worktree checkouts return `PROJECT_DIR` verbatim — behavior is byte-identical as claimed.
- Migration + gitignore keying stays consistent with the redirect; no stray `$PROJECT_DIR/.remember` derivations elsewhere.

The fail-safe design (common-dir ≠ git-dir + main-checkout-is-worktree guard) is exactly right, and the tests using real `git worktree` layouts are appreciated.

One minor, non-blocking follow-up (in `session-start-hook.sh`, not touched by this PR): keying `REMEMBER_DIR` to the main checkout makes `REMEMBER_ROOT = dirname(REMEMBER_DIR)` differ from `PROJECT_DIR` in a **legacy-mode worktree**. That's what correctly fires the handoff hint (line ~105) — but it also flips the identity-tier branch at line ~88, and the comment at line ~84 ("In legacy mode REMEMBER_ROOT == PROJECT_DIR") is no longer true for worktrees. Real impact is small (needs a root-level `identity.md`), so it's fine to land this and clean that up separately.

LGTM. 👍

PR #130, opened 2026-07-16
fix: register the UserPromptSubmit hook in the plugin manifest

mlalewicz:
## Problem

`scripts/user-prompt-hook.sh` ships with the plugin, and the README documents it in two places — the manual-install `settings.json` snippet and the hooks table ("The plugin registers **three** Claude Code hooks") — but `hooks/hooks.json` only registers **two**: `SessionStart` and `PostToolUse`.

For everyone installing via the plugin marketplace, that makes two documented features dead code:

- the per-prompt timestamp injection (`[HH:MM TZ — username]`), so the agent never knows the current time mid-session;
- the `after_user_prompt` lifecycle dispatch to `hooks.d/` listeners.

Manual installs per the README snippet are unaffected, since that snippet wires all three hooks correctly.

Easy to verify: install the plugin from the marketplace, submit any prompt — no timestamp is injected and `user-prompt-hook.sh` never runs.

## Fix

- Register `UserPromptSubmit` in `hooks/hooks.json`, mirroring the existing entries (same `bash \"${CLAUDE_PLUGIN_ROOT}/scripts/...\"` form, so it's covered by the existing quoting/PowerShell lint tests automatically).
- Add `test_every_shipped_hook_script_is_wired` to `tests/test_hooks_json.py`: every `scripts/*-hook.sh` must be referenced by a `hooks.json` command. Red on the current manifest, green with the fix — guards against this class of gap recurring.
- CHANGELOG entry under `[Unreleased]`.

## Testing

- `pytest tests/test_hooks_json.py` — 7 passed (pwsh/Git-Bash legs skipped locally; CI covers them).
- New test verified red against the unfixed manifest, green after.
- Full suite: 501 passed, 43 skipped.
- Running the patched manifest daily in my own setup — timestamp injection fires on every prompt.


fdaviddpt:
Merged — thanks, this closed a real gap: the script shipped and the README documented it, but marketplace installs never had it wired.

One follow-up worth confirming rather than assuming: a user who installed via the plugin **and** also wired the hook manually from the README snippet now has it registered twice, so `user-prompt-hook.sh` runs twice per prompt. The visible effect is a duplicated timestamp injection — harmless, but it means the README snippet and the plugin manifest now overlap. Either the README should note that manual wiring is only for non-plugin installs, or the hook should be idempotent per prompt. Not a blocker for this PR.

---
*Reviewed by Max (AI dev partner at Digital Process Tools).*

PR #137, opened 2026-07-24
fix: don't crash the nested Haiku session on unresolvable project root

lucasrodriggs-tech:
## Summary
- `haiku.py`'s `call_haiku()` spawns a nested `claude -p` process to summarize each save. Since plugins are installed globally, that nested session also loads this plugin and fires its own `SessionStart` hook — an unintended recursive activation.
- That nested session runs with `cwd=tempdir` and no real project, so `CLAUDE_PROJECT_DIR` is unset for it. `resolve-paths.sh` correctly detects it can't guess a project root from a marketplace-cache layout, but signals that with a bare `exit 1` inside a script that is always `source`d — which kills the *caller's* whole process, even though `session-start-hook.sh` documents "EXIT CODES: 0 Always (hook must not block session startup)".
- Claude Code then reports the whole nested `claude -p` invocation as failed, which `call_haiku()` surfaces as `claude exited 1`, so the outer save silently never reaches `now.md`.

## Fix
- `resolve-paths.sh` now signals failure via `return 1` (correct for a sourced library) instead of `exit 1`.
- `session-start-hook.sh`, `post-tool-hook.sh`, `user-prompt-hook.sh` — the three entrypoints documented to never block the host session — now guard the source call with `|| exit 0`, matching their own contract.
- `save-session.sh` / `run-consolidation.sh` needed no change: they already run under `set -e`, so a failed `return 1` still correctly aborts them.

## Test plan
- [x] Reproduced the crash: forced a save with `CLAUDE_PROJECT_DIR` unset in the nested hook context, confirmed `[resolve] FATAL` + `claude exited 1` in logs.
- [x] Verified `source resolve-paths.sh` now returns 1 without killing the sourcing shell.
- [x] Verified `session-start-hook.sh` exits 0 cleanly instead of crashing when project root can't be resolved.
- [x] Ran a real `save-session.sh --force` end-to-end after the fix: no more `[resolve] FATAL`, Haiku call succeeds, `now.md` gets a real appended entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **A nested `claude -p` session crashed on every hook** ([#137](https://github.com/Digital-Process-Tools/claude-remember/pull/137)) — `scripts/resolve-paths.sh` is always *sourced*, but signalled failure with a bare `exit 1`, which terminates the **caller's** whole process. The three Claude Code hooks are documented "EXIT CODES: 0 Always", and the one caller most likely to fail resolution is the plugin's own nested Haiku session (it runs with `cwd` in a temp dir and no `CLAUDE_PROJECT_DIR`, so it is not a project at all) — so the plugin crashed the very session it spawned. Resolution failure now stays **loud by default** (`exit 1`, unchanged for every worker script and for any caller that forgets to check), while a caller that must never take its host process down opts in with `REMEMBER_PATHS_SOFT_FAIL=1` and gets `return 1` to handle itself. Only the three hooks opt in, and they still report FATAL on stderr, which `hooks.json` redirects into `hook-errors.log` — a failed hook is silent to the session, never silent to the logs. Diagnosed and fixed by [@lucasrodriggs-tech](https://github.com/lucasrodriggs-tech).
+
 ## [0.8.5] — Ship the 0.8.4-era fixes: manifest bump, worktree safety, fork storm
 
 ### Fixed
--- a/scripts/post-tool-hook.sh
+++ b/scripts/post-tool-hook.sh
@@ -28,11 +28,11 @@
 # ============================================================================
 
 # --- Resolve paths ---
-# resolve-paths.sh signals failure via `return`, not `exit` — see the comment
-# in session-start-hook.sh. This hook must never block the agent, so a
-# resolution failure (e.g. a nested/headless session with no
-# CLAUDE_PROJECT_DIR) is a silent no-op, not a crash.
-source "$(dirname "$0")/resolve-paths.sh" || exit 0
+# Opt into resolve-paths.sh's soft-failure mode — see the comment in
+# session-start-hook.sh. This hook must never block the agent, so a resolution
+# failure (e.g. a nested/headless session with no CLAUDE_PROJECT_DIR) is a
+# silent no-op, not a crash.
+REMEMBER_PATHS_SOFT_FAIL=1 source "$(dirname "$0")/resolve-paths.sh" || exit 0
 source "$(dirname "$0")/detect-tools.sh"
 source "$(dirname "$0")/bootstrap-dirs.sh"
 PLUGIN_ROOT="$PIPELINE_DIR"
--- a/scripts/post-tool-hook.sh
+++ b/scripts/post-tool-hook.sh
@@ -28,7 +28,11 @@
 # ============================================================================
 
 # --- Resolve paths ---
-source "$(dirname "$0")/resolve-paths.sh"
+# resolve-paths.sh signals failure via `return`, not `exit` — see the comment
+# in session-start-hook.sh. This hook must never block the agent, so a
+# resolution failure (e.g. a nested/headless session with no
+# CLAUDE_PROJECT_DIR) is a silent no-op, not a crash.
+source "$(dirname "$0")/resolve-paths.sh" || exit 0
 source "$(dirname "$0")/detect-tools.sh"
 source "$(dirname "$0")/bootstrap-dirs.sh"
 PLUGIN_ROOT="$PIPELINE_DIR"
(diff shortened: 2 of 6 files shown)


fdaviddpt:
Thanks for this — the diagnosis is right and the comments you added to `resolve-paths.sh` explain the `return`-vs-`exit` contract better than the original did.

I checked the thing most likely to break with this change: callers that *relied* on the sourced `exit 1` to hard-stop. Both `save-session.sh` and `run-consolidation.sh` place `set -e` above their `source`, so a `return 1` still aborts them — behaviour preserved. `run-consolidation.sh` isn't touched by this PR and is the one I'd have expected to regress, so that's worth stating explicitly somewhere.

Two things before this goes in:

**1. It changes failure semantics for three hooks and adds no test.** The natural one: run each hook with `CLAUDE_PROJECT_DIR` unset and `CLAUDE_PLUGIN_ROOT` pointing nowhere useful, assert exit 0 and no stdout. `tests/test_tmpdir_cleanup.py` (from #134) is a good model for invoking the real scripts under an isolated environment. This matters more than usual here because CI on this repo needs maintainer approval for outside contributions, so PRs land with less automated cover than you'd expect.

**2. The failure is now quiet in the place it's most likely to happen.** `resolve-paths.sh` still echoes `FATAL: …` to stderr before returning, so the signal exists — but `post-tool-hook.sh` runs where stderr is usually discarded. Net effect: a genuinely misconfigured project stops saving memory silently instead of failing loudly, which is the opposite of the old behaviour.

For the nested-session case you're fixing, silent is exactly right. For a real project it's a regression in observability. Would you consider distinguishing the two — e.g. no-op silently when `CLAUDE_PROJECT_DIR` is unset (the nested `claude -p` signature), but keep a visible complaint when it's set and resolution still fails? That keeps the fix and preserves the "something is wrong with your install" signal.

Happy to merge once there's a test; the rest is solid.

---
*Reviewed by Max (AI dev partner at Digital Process Tools) — happy to answer follow-ups here.*

fdaviddpt:
Follow-up: I approved the workflow run (outside-contributor PRs on this repo sit at `action_required` until a maintainer does, which is why this PR had no CI until now). It comes back **8 failures out of 12**, and they're not flakes — they're existing tests that encode the contract this PR changes:

```
tests/test_path_resolution.py:302
  AssertionError: Should fail when marketplace install has no CLAUDE_PROJECT_DIR
  assert 0 != 0

tests/test_path_resolution.py:677
  AssertionError: Marketplace install without CLAUDE_PROJECT_DIR should fail
```

`test_marketplace_without_env_vars_fails_loud` sources `resolve-paths.sh` from a wrapper that does *not* check the return value, and asserts the process exits non-zero. Under `exit 1` it did; under `return 1` the wrapper carries on and exits 0.

So this isn't a matter of updating a stale assertion — it surfaces the actual design question:

- **`exit 1`** is fail-safe by default. Any caller, including one written later by someone who never read the contract, stops. Its cost is the bug you're fixing: it kills hooks that are documented to never block.
- **`return 1`** lets hooks no-op, but is fail-open by default. A caller that forgets `|| …` silently proceeds with `PROJECT_DIR` and `PIPELINE_DIR` unset — which is precisely the "silently compute wrong paths" outcome that test was written to prevent.

A middle path that keeps both properties: have `resolve-paths.sh` set an explicit sentinel (`RESOLVE_PATHS_OK=1` on success, unset/0 on failure) alongside `return 1`, and have `bootstrap-dirs.sh` — which every caller sources immediately afterwards — hard-fail when the sentinel is absent. Callers that genuinely want to no-op (your three hooks) keep their `|| exit 0` and never reach it; a caller that forgets still fails loudly instead of running with empty paths. The two failing tests would then need their wrapper updated to reflect the new contract, with the sentinel keeping their *intent* — fail loud on a real misconfiguration — intact.

Whichever direction you prefer, the tests need to change deliberately as part of this PR rather than as a follow-up, since they're the written record of the old contract.

One more thing worth separating: a nested `claude -p` session and a broken marketplace install both present as "no `CLAUDE_PROJECT_DIR`", so that variable alone can't discriminate the case you want to silence from the case that should shout. If there's a signal unique to the nested pipeline call (an env var the pipeline sets when it invokes Haiku, say), gating on that would let you fix the crash without loosening the contract at all — probably the cleanest version of this change.

---
*Reviewed by Max (AI dev partner at Digital Process Tools) — happy to answer follow-ups here.*

fdaviddpt:
Thanks for this — the diagnosis is exactly right, and it's a good catch. A bare `exit` in a sourced file taking down the caller is the kind of bug that reads as correct until the caller is a hook documented "EXIT CODES: 0 Always". Your nested-session repro checks out against `_child_env()` in `pipeline/haiku.py`: temp cwd, no `CLAUDE_PROJECT_DIR`, so the plugin was crashing the very session it spawned.

Rather than send it back for a round trip, I pushed a commit onto your branch (maintainer edits were on) and rebased it on current main. Two things it changes:

**1. Five existing tests in `tests/test_path_resolution.py` were red.** They encode the contract "sourcing resolve-paths.sh with bad env must fail loud", which the switch to `return 1` relaxed for *everyone*, not just the hooks. Worth seeing why that matters beyond the tests: `_make_path_probe()` builds a plain wrapper with no `set -e` and no guard. Under `return 1` it sails past the failure and continues with empty `PROJECT_DIR`/`PIPELINE_DIR` — which is the silent write-to-the-wrong-directory failure we'd just spent the morning fixing in #138. Your current callers are all safe; the next one wouldn't be.

**2. So the default is inverted instead of relaxed.** `resolve-paths.sh` still exits the caller on failure. A caller that must never take its host process down sets `REMEMBER_PATHS_SOFT_FAIL=1` before sourcing and gets `return 1` to handle itself — only the three hooks do, paired with your `|| exit 0`. Same fix for your crash, but a caller who forgets to check still fails safely.

Two of the five tests went green untouched after that (the contract they pin is intact). The other three genuinely needed updating — they asserted hooks exit non-zero, which is the thing your PR fixes; they now assert the hooks exit 0, emit nothing, and *still* print FATAL to stderr so the `hooks.json` redirect keeps capturing it in `hook-errors.log`. Silent to the session, never silent to the logs.

12/12 green across the matrix. Merging now, shipping in the next release — your commit and authorship stay on it. One note for next time: run `pytest tests/test_path_resolution.py` locally, it's the file that encodes this exact contract and would have shown you the five in about 40 seconds.

Nice work — genuinely useful fix.

— Max

Issue #126, opened 2026-07-12
`consolidate`: wrapping code fence in Haiku output produces a doubled `# Recent` header in recent.md


merpuya:
## What happens

`recent.md` (and `archive.md`) periodically get written with a duplicated header and an orphaned code fence, e.g.:

~~~
# Recent

```

# Recent

## 2026-07-02
Committed CF skill pack.
~~~

The stray ` ``` ` and the second `# Recent` corrupt the file that `/resume`-style tooling and the SessionStart banner read, so downstream consumers mis-parse "recent" state.

## Root cause

`parse_consolidation_response()` in `pipeline/consolidate.py` normalizes the header like this (lines 185–186 on current `main`, v0.8.4):

```python
recent = parts[0].replace("===RECENT===", "").strip()
...
if recent and not recent.startswith("# Recent"):
    recent = "# Recent\n\n" + recent
```

`.strip()` removes whitespace but **not** a Markdown code fence. When Haiku returns the `===RECENT===` body wrapped in a fence (```` ```markdown … ``` ````) or emits a stray leading ```` ``` ```` line, the body starts with the fence line — so `startswith("# Recent")` is `False`, and a **second** `# Recent` header is prepended. The opening fence is left orphaned (truncated output often drops the closing fence).

It runs for every buffer, so the artifact is systemic rather than a one-off. First seen on v0.8.3; verified still present on `main` (v0.8.4).

## Fix

Strip a wrapping fence (opening always; closing only if present) from each section body *before* the header-ensure check: add a `_strip_wrapping_fence()` helper and route `recent` and `archive` through it in `parse_consolidation_response()`. Two regression tests cover the stray-leading and full-`` ```markdown `` cases. With the patch below applied to v0.8.3: 20 passed (18 existing + 2 new).

<details>
<summary>Patch (applies to v0.8.3; also applies to current main)</summary>

```diff
--- a/pipeline/consolidate.py
+++ b/pipeline/consolidate.py
@@ -31,7 +31,38 @@
 # A real memory entry header: "## HH:MM", "## Week of ...", or "## YYYY-MM-DD".
 _ENTRY_HEADER = re.compile(r"(?m)^## (\d{2}:\d{2}|Week of |\d{4}-\d{2}-\d{2})")
 
+# A Markdown code fence the model sometimes wraps a whole section body in:
+# an opening ``` (optionally ```markdown) line, and a matching closing ```.
[...]

PR #154, opened 2026-07-25
fix: only strip a code fence that is genuinely a wrapper

fdaviddpt:
Follow-up to the #126 fix shipped in **v0.8.6**. A post-release adversarial review found it incomplete in both directions — this is a live regression in a published release.

## It corrupted legitimate content

The shipped code strips a leading fence and then *any* trailing fence, without checking the two are a pair. Verified against the released function:

```
IN : '```markdown\n# Recent\n\n## 12:00\nRan:\n\n```bash\nls -la\n```'
OUT: '# Recent\n\n## 12:00\nRan:\n\n```bash\nls -la'      <- block never closes

IN : '```bash\nls -la\n```'
OUT: 'ls -la'                                            <- fences deleted
```

The first is the realistic one, and the old docstring anticipated exactly that input: a wrap whose closer was truncated, over a body ending in a code sample. Result is an unterminated code block — everything after it renders as code in `recent.md`.

## And it still shipped the original artifact

When the model wraps the **whole** response, the closing fence sits at the end of the *archive* section, which has no opener of its own — so the per-section strip can't see it and the orphan ` ``` ` reaches `archive.md`. That is the shape #126 was reported with.

## Fix

Recognise a wrapper rather than assume one:

- opener must be bare or document-tagged (`markdown`/`md`/`text`/`txt`/`plaintext`) — ` ```bash ` opens content
- closer removed only when the fences in between are balanced, so inner blocks keep their terminator
- unbalanced (truncated output) → strip the opener only
- whole-response wrapper stripped **before** the delimiter split

## Tests

Five new tests; **three fail against the shipped 0.8.6 code**. 542 passed.

Not merging until this has had an independent review.


fdaviddpt:
Marking this draft. Third attempt at this function, third rejection — and this one is the clearest signal yet: a differential sweep against shipped v0.8.6 found **180 inputs where this branch produces an artifact that 0.8.6 does not**.

The killer is the most ordinary model output there is:

```
```markdown
# Recent
...
```

Let me know if you want changes.
```

A trailing sentence after the closer means the closer is no longer the last line, so my "it's only a wrapper if the closer is last" rule declines to strip — and the surviving wrapper reproduces #126 verbatim, doubled header and all. 0.8.6 handles that input cleanly.

Also unresolved: the ambiguity guard is blind to any info string with a space or quote (```bash -x "y"`, ```json {"a":1}`), so it still eats a real terminator; `_strip_wrapping_fence` is not idempotent while `parse_consolidation_response` now applies it twice, which destroys a legitimate ```text block inside a ````markdown wrapper; and a whole-body ```text log sample is still unwrapped as if it were prose.

**The approach is wrong, not just the details.** Three attempts have all tried to *recognise and remove a wrapper*, which means parsing Markdown correctly — and each heuristic dies to a shape the previous one didn't consider.

The artifact in #126 is narrower than that: a doubled `# Recent` header, and an orphan fence. Neither requires understanding the fence structure:

1. **Doubled header** — make the `startswith("# Recent")` check skip leading fence lines instead of mutating the body. No content is touched, so no shape can corrupt it.
2. **Orphan fence** — remove a trailing bare fence only when the section's fence count is *odd*, which is exactly the definition of an orphan.

That fixes the reported bug without ever needing to decide what is or isn't a wrapper, and it cannot unbalance content because it only ever removes an already-unbalanced fence.

Leaving this branch up for the record. Next session picks it up with that design.

PR #155, opened 2026-07-25
fix(consolidate): strip the orphaned closing fence from the archive half (#126)

rmerk:
## The other half of #126

`_strip_wrapping_fence` removes a closing fence only when the **same** section also started with an opening one:

```python
m = _OPEN_FENCE.match(t)
if m:
    t = t[m.end():]
    t = _CLOSE_FENCE.sub("", t)   # ← only reachable when this section had an opener
```

That holds when the model fences a single section body. It does not hold when the model fences the **whole** `===RECENT===`/`===ARCHIVE===` envelope — which is what the output-format block in `prompts/consolidate-staging.prompt.txt` asks for, since it shows the envelope inside a fence under "Return EXACTLY this structure".

Then the opener lands in the recent half and the closer in the archive half. The archive half has no opener of its own, so the gated strip never fires and the ``` survives into `archive.md`.

So #126's doubled `# Recent` header is fixed, but the *orphaned-fence artifact* the docstring names still lands — just on the archive side.

### Reproduction

```python
>>> parse_consolidation_response(
...     "```\n===RECENT===\n# Recent\n\nx\n\n===ARCHIVE===\n# Archive\n\ny\n```")
('# Recent\n\nx', '# Archive\n\ny\n```')
                                 ^^^ orphan
```

Observed in the wild on 0.8.3 and reproduced against 0.8.6: `recent.md` opened with an unclosed fence and `archive.md` ended with the matching closer. It matters mildly but persistently — `archive.md` is re-fed as the next consolidation's input and is injected into every session's context.

## The fix

Decide on **balance** rather than on the opener: an odd number of fence delimiters means one is unmatched, so the trailing fence is the orphan and is removed; an even number means they pair up, so a section legitimately ending in a fenced code block keeps it.

An unconditional `_CLOSE_FENCE.sub` would have been the smaller diff, but it silently corrupts a section whose body genuinely ends in a code block — `test_parse_keeps_balanced_code_block_at_end_of_section` covers that case.

## Tests

Two added to `tests/test_consolidate.py`:

- `test_parse_strips_orphaned_closing_fence_from_archive` — fails on `main`, passes with the fix
- `test_parse_keeps_balanced_code_block_at_end_of_section` — guards the fix against over-stripping

Full suite: **538 passed, 42 skipped**, coverage 98.21%. Existing fence tests (`test_parse_strips_stray_leading_fence_no_double_header`, `test_parse_strips_wrapping_markdown_fence`) unchanged and green.

---

Found while diagnosing an unrelated 8-day outage on 0.8.3 where every memory write failed silently — thank you for the `_failure_detail()` change in 0.8.6, which is exactly what would have surfaced it.

### DIFF
--- a/pipeline/consolidate.py
+++ b/pipeline/consolidate.py
@@ -49,6 +49,17 @@ def _strip_wrapping_fence(text: str) -> str:
     Strip the opening fence whenever present, and the closing fence only if
     present (truncated model output can drop it), before the header check runs.
 
+    The closing fence is NOT gated on this section having had an opener. When
+    the model fences the whole ===RECENT===/===ARCHIVE=== envelope rather than
+    one section, the opener lands in the FIRST section and the closer in the
+    LAST — so the archive half inherits an unmatched ``` with no opener of its
+    own, and an opener-gated strip would leave it to land in archive.md.
+
+    Balance decides it: an odd number of fence delimiters means one is
+    unmatched, so the trailing fence is the orphan and is removed. An even
+    number means they pair up, so a section legitimately ending in a fenced
+    code block keeps it.
+
     Args:
         text: A section body, before header normalization.
 
@@ -59,6 +70,7 @@ def _strip_wrapping_fence(text: str) -> str:
     m = _OPEN_FENCE.match(t)
     if m:
         t = t[m.end():]
+    if sum(1 for ln in t.split("\n") if ln.lstrip().startswith("```")) % 2:
         t = _CLOSE_FENCE.sub("", t)
     return t.strip()
 
--- a/tests/test_consolidate.py
+++ b/tests/test_consolidate.py
@@ -300,3 +300,32 @@ def test_parse_strips_wrapping_markdown_fence():
     assert "```" not in recent
     assert archive.startswith("# Archive")
     assert "```" not in archive
+
+
+def test_parse_strips_orphaned_closing_fence_from_archive():
+    """A fence wrapping the WHOLE envelope leaves its closer in the ARCHIVE half.
+
+    The opener lands before ===RECENT=== and the closer after the archive body,
+    so the archive section inherits an unmatched ``` with no opener of its own —
+    _strip_wrapping_fence's opener-gated close-strip never fires and the fence
+    survives into archive.md (the other half of issue #126).
+    """
+    recent, archive = parse_consolidation_response(
+        "```\n"
+        "===RECENT===\n# Recent\n\n## 2026-07-20\nDid a thing.\n\n"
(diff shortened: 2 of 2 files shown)


fdaviddpt:
Thanks for this — the diagnosis is exactly right, and the "opener in the recent half, closer in the archive half" framing is the clearest statement of that failure mode I've seen.

Timing was unkind: #154 landed a few hours ago and rewrote `_strip_wrapping_fence` end to end, so this won't apply as-is — the `_OPEN_FENCE`/`_CLOSE_FENCE` gate it patches is gone. Your repro passes on current `main`:

```python
>>> parse_consolidation_response(
...     "```\n===RECENT===\n# Recent\n\nx\n\n===ARCHIVE===\n# Archive\n\ny\n```")
('# Recent\n\nx', '# Archive\n\ny')
```

and so do both your tests, unchanged.

One note on the approach, because we tried it too. Parity was the first thing I shipped in that chain, and I removed it two commits later: a section containing a nested longer fence counts three fence-ish lines, so parity reads "odd → strip the trailing one" and uncloses the block.

```
# Archive

Fence example:

````
```
````
```

What replaced it walks the body tracking nesting depth, and only treats a fence as the wrapper's closer when it lands on the last line at depth zero. When no closer arrives, what is left *open* decides: a dangling fence that could have closed the wrapper means the leading fence enclosed part of the content, while a dangling ` ```bash ` or `~~~` could never have closed it and so cannot keep the wrapper alive.

Your bare-fence envelope fixture covers a shape our suite didn't — ours wraps with ` ```markdown `, and a bare ` ``` ` opener is the more ambiguous case of the two. I'd like to add `test_parse_strips_orphaned_closing_fence_from_archive` with credit to you, if that's alright.

And thank you for the note about `_failure_detail()` — an 8-day silent outage is exactly what it was built for, and it's good to hear it did its job.

— Max

fdaviddpt:
Closing this as superseded by #154, which is now on `main` — your repro and both your tests pass there unchanged.

To be clear about why, since the timing was pure bad luck rather than anything wrong with the PR: you identified the right bug and described it better than our own docstring did. #154 happened to land the same afternoon and rewrote the whole function, so there is no longer a `_OPEN_FENCE`/`_CLOSE_FENCE` gate for this to patch.

The offer stands on your bare-fence envelope fixture — ours wraps with ` ```markdown `, and a bare ` ``` ` opener is the genuinely harder case of the two. Happy to add it with credit to you if you'd like; just say so here and I'll open it.

Thanks for the report, and for the `_failure_detail()` note — an 8-day silent outage is exactly what it was built for.

— Max

Issue #169, opened 2026-07-26
CLAUDE_CONFIG_DIR is not path-normalised on Windows, unlike PROJECT_DIR


fdaviddpt:
## Description

`claude_projects_dir()` (added in #166) concatenates `CLAUDE_CONFIG_DIR` with a forward slash and does no path normalisation:

```bash
printf '%s/projects' "${CLAUDE_CONFIG_DIR:-$HOME/.claude}"
```

`scripts/resolve-paths.sh` deliberately normalises `PROJECT_DIR` from Git Bash's POSIX form (`/c/Users/x`) to native Windows form (`C:\Users\x`) so the slug matches the directory Claude Code actually created. There is no equivalent handling for `CLAUDE_CONFIG_DIR`.

So on Windows, a natively-expressed value inherited from the environment produces a mixed-separator path:

```
C:\Users\x\.claude-alt/projects/<slug>
```

Whether MSYS's path layer resolves that is genuinely unknown — it may well be fine, since MSYS translates both forms in syscalls.

## Why it is unverified

Raised during review of #166 and not closed, because there is no Windows machine here to test on. The one test that would exercise it, `test_the_hook_finds_the_transcript_in_the_relocated_tree`, is `skipif(sys.platform == "win32")` in line with the repo's convention for bash-subprocess tests — so this path has **zero coverage on the only platform where it matters**.

## Why it is worth closing properly

The failure mode is the one this project keeps hitting: the session directory is not found, `ls` returns nothing, and the pipeline no-ops in silence. #166 is the fourth report of that shape. Since #144 the hook at least logs a warning when the session directory is missing, so it would not be *completely* silent now — but a Windows user with `CLAUDE_CONFIG_DIR` set would still get no memory and only a log line to go on.

## Suggested approach

- Decide whether `CLAUDE_CONFIG_DIR` needs the same `cygpath` treatment `PROJECT_DIR` gets in `resolve-paths.sh`, or whether MSYS's own translation makes it moot — this needs someone on Windows, or a Windows CI job that actually exercises it.
- If the tests stay skipped on `win32`, that is a defensible convention, but it means this class can only ever be caught by a user. Worth considering one Windows-runner test that builds a relocated tree and asserts the transcript is found.

Issue #166, opened 2026-07-26
Save pipeline silently no-ops when CLAUDE_CONFIG_DIR is set


samoilovartem:
## Summary

The plugin resolves Claude Code's session directory as `$HOME/.claude/projects/<slug>/`. But Claude Code relocates that entire tree — `projects/` included — when `CLAUDE_CONFIG_DIR` is set, which people use to run a separate account per project. The plugin never reads that variable, so it looks in a directory with no current transcripts and the whole save pipeline no-ops. Silently.

This is the same failure shape already documented for the Windows slug mismatch in `scripts/resolve-paths.sh`:

> The plugin's `ls $SESSION_DIR/*.jsonl` then returns nothing and the entire save pipeline silently no-ops.

## Impact

In my case: 5 days, ~2000 `PostToolUse` hook invocations, zero saves. No `now.md`, no `today-*.md`, `logs/autonomous/` empty, `tmp/` empty. Nothing in the logs suggests a problem — the hook writes its env line and exits 0 every time.

The concrete path: `post-tool-hook.sh` computes `DELTA = CURRENT_LINES - LAST_LINE` from whatever stale `.jsonl` happens to be newest in the wrong directory. Mine had 16 lines and `delta_lines_trigger` is 50, so `save-session.sh` was never forked.

There is a worse variant: if the wrong directory holds an old transcript with more than `delta_lines_trigger` lines, the pipeline will happily summarize *that stale session* into memory instead of the real one.

## Affected sites (v0.8.6)

- `scripts/post-tool-hook.sh:45`
- `scripts/save-session.sh:100`
- `scripts/session-start-hook.sh:73`
- `pipeline/extract.py:48`

## Reproduce

1. `export CLAUDE_CONFIG_DIR=~/.claude-alt` (e.g. via direnv) and start Claude Code in that project
2. Work long enough to pass `thresholds.delta_lines_trigger`
3. `.remember/logs/autonomous/` stays empty and no `now.md` is ever written

## Suggested fix

Follow the variable the same way Claude Code itself does:

```bash
# the three shell sites
"${CLAUDE_CONFIG_DIR:-$HOME/.claude}/projects/..."
```

```python
# pipeline/extract.py::_session_dir
config_dir = os.environ.get("CLAUDE_CONFIG_DIR") or (home + "/.claude")
return config_dir + "/projects/" + slug
```

[...]

Issue #158, opened 2026-07-25
Duplicate naive session_dir_slug in lib-memory-dir.sh carries the pre-#144 bugs for user-prompt-hook


fdaviddpt:
## Description

There are two implementations of the session-directory slug, and only one of them is correct.

`scripts/lib-memory-dir.sh` carries an inline fallback:

```bash
# Guard: session_dir_slug may not be defined if detect-tools.sh was
# not sourced yet (e.g. log.sh sourced directly in tests). Define a
# minimal inline fallback so the slug is never silently empty.
if ! type session_dir_slug >/dev/null 2>&1; then
    session_dir_slug() {
        local _p="$1"
        command -v cygpath >/dev/null 2>&1 && _p=$(cygpath -m "$_p")
        echo "$_p" | sed 's/[^a-zA-Z0-9]/-/g'
    }
fi
```

That is the pre-#144 implementation, carrying every bug #156 just fixed: no `LC_ALL`, so byte-wise on Git Bash/MSYS; no UTF-16 handling, so astral characters get one dash instead of two; and `echo | sed`, so an embedded newline never reaches the substitution.

It is not dead code. `scripts/user-prompt-hook.sh` sources `bootstrap-dirs.sh` **without** sourcing `detect-tools.sh` first, so for that hook the fallback is what runs.

## Impact

Only when `data_dir` uses the `{slug}` placeholder (external storage mode — see README "External storage mode"), which is where the slug feeds `REMEMBER_DIR` rather than the transcript lookup.

In that configuration, on a project whose path contains non-ASCII, `user-prompt-hook.sh` resolves a **different** `REMEMBER_DIR` than every other hook. So one hook reads and writes memory in one directory while the rest use another — a split-brain memory rather than a clean failure, which is harder to notice than the silent no-op of #144.

Also, `cygpath -m` here versus `cygpath -w` plus drive-letter lowercasing in the real function is a second, independent divergence on Windows.

## Suggested fix

Extract the slug into its own small file — `scripts/lib-slug.sh` — holding `session_dir_slug` and its source-time sed construction, and source that from both `detect-tools.sh` and `lib-memory-dir.sh`. One implementation, no drift.

Sourcing `detect-tools.sh` from `lib-memory-dir.sh` instead is **not** a safe shortcut: `detect-tools.sh` runs Python detection at source time and calls `exit 1` when it finds none, which would take down whatever sourced it. That is presumably why the inline duplicate exists in the first place.

[...]

PR #171, opened 2026-07-26
fix(hooks): honor CLAUDE_CONFIG_DIR instead of hardcoding $HOME/.claude

Jutiphan:
## The bug

Claude Code supports relocating its entire config root via the
`CLAUDE_CONFIG_DIR` environment variable — used, for example, to run a
second account on one machine (`CLAUDE_CONFIG_DIR=~/.claude-max`). When
set, session transcripts live under `$CLAUDE_CONFIG_DIR/projects/<slug>/`,
not `$HOME/.claude/projects/<slug>/`.

Four sites in the plugin hardcode the default path outright, ignoring
`CLAUDE_CONFIG_DIR` entirely:

- `scripts/save-session.sh` (`SESSION_DIR_PATH`, used for session-id
  auto-detection)
- `scripts/session-start-hook.sh` (`SESSIONS_DIR`, used for missed-session
  recovery)
- `scripts/post-tool-hook.sh` (`SESSION_DIR`, used for the delta-based
  save trigger)
- `pipeline/extract.py:_session_dir` (used by every Python-side session
  lookup, including the actual extraction itself)

## Why it matters

Under a non-default `CLAUDE_CONFIG_DIR`, every one of these looks in the
*wrong account's* `~/.claude/projects/`. Depending on what's there, this
fails one of two ways:

- Nothing under the default path for this project → every hook silently
  no-ops (finds no transcript, does nothing) — memory simply never gets
  written for that account, with no error anywhere.
- The *other* account happens to have session data for a directory of the
  same slug (the same project, opened from both accounts) → this account's
  hooks read and summarize **the other account's transcripts** into this
  account's memory.

## The fix

`${CLAUDE_CONFIG_DIR:-$HOME/.claude}` in shell (all three scripts), and
the equivalent in Python:

```python
config_dir = os.environ.get("CLAUDE_CONFIG_DIR") or (home + "/.claude")
return config_dir + "/projects/" + slug
```

Bash's `:-` (colon-dash) already treats *unset or null* as "use the
default", so `${CLAUDE_CONFIG_DIR:-$HOME/.claude}` handles an
empty-but-exported `CLAUDE_CONFIG_DIR=""` correctly with no extra logic.
The Python side needed `or`, not a bare dict lookup, for the same reason
— a naive `os.environ.get("CLAUDE_CONFIG_DIR") + "/projects"` would
produce a path starting `/projects` for an empty value.

## Test evidence

New file: `tests/test_claude_config_dir.py`:
- Three unit tests against `pipeline.extract._session_dir` (set, unset,
  empty-string `CLAUDE_CONFIG_DIR`).
- One end-to-end test running the real `save-session.sh --dry` with a
  session transcript planted under `$CLAUDE_CONFIG_DIR/projects/<slug>/`
  and a decoy transcript planted under the default
  `$HOME/.claude/projects/<slug>/` — asserts the script's session-id
  auto-detection finds the real one, not the decoy.

Before (unpatched `main`, HEAD `d9480ec`):
```
FAILED tests/test_claude_config_dir.py::TestExtractSessionDirHonoursConfigDir::test_uses_claude_config_dir_when_set
FAILED tests/test_claude_config_dir.py::TestShellHooksHonourConfigDir::test_save_session_finds_transcript_under_config_dir
2 failed, 2 passed in 0.36s
```
(The 2 passes are the unset and empty-string fallback cases, which already
matched the default; included to pin the invariant down explicitly, not
to demonstrate the bug.)

After (patched):
```
4 passed in 0.66s
```

### Regression suite

This machine has no `jq`, so `main`'s own `pytest tests/ -q` baseline
already fails a batch of tests unrelated to this change:

- Baseline (`main`, HEAD `d9480ec`): `38 failed, 625 passed, 58 skipped`
- Patched: `38 failed, 629 passed, 58 skipped` (identical pre-existing
  failures, +4 for this patch's new tests)

## Backward compatibility

None needed. Every existing installation runs with `CLAUDE_CONFIG_DIR`
unset, which is exactly the fallback branch — byte-for-byte identical
behavior to `main`.

---
This PR is independent of the other four bug-fix PRs I'm submitting
alongside it and can be reviewed/merged in any order.

### DIFF
--- a/pipeline/extract.py
+++ b/pipeline/extract.py
@@ -45,7 +45,15 @@ def _session_dir(project_dir: str) -> str:
     # Honor HOME explicitly so test fixtures patching only HOME also work on Windows
     # (where os.path.expanduser defaults to USERPROFILE, ignoring HOME).
     home = os.environ.get("HOME") or os.path.expanduser("~")
-    return home + "/.claude/projects/" + slug
+    # Claude Code honors CLAUDE_CONFIG_DIR to relocate its whole config root
+    # (e.g. a second account on one machine: CLAUDE_CONFIG_DIR=~/.claude-max)
+    # — sessions then live under "$CLAUDE_CONFIG_DIR/projects/", not
+    # "~/.claude/projects/". An unset *or empty* value must fall back to the
+    # default; "or" (not a bare os.environ["..."] lookup) covers both, unlike
+    # a naive os.environ.get(...) + "/projects" which would yield a path
+    # starting "/projects" for CLAUDE_CONFIG_DIR="".
+    config_dir = os.environ.get("CLAUDE_CONFIG_DIR") or (home + "/.claude")
+    return config_dir + "/projects/" + slug
 
 
 def _is_line_number(value: object) -> bool:
--- a/scripts/post-tool-hook.sh
+++ b/scripts/post-tool-hook.sh
@@ -42,7 +42,7 @@ log "hook" "post-tool: PROJECT_DIR=$PROJECT_DIR PIPELINE_DIR=$PIPELINE_DIR PYTHO
 SAVE_SCRIPT="$PLUGIN_ROOT/scripts/save-session.sh"
 LAST_SAVE_FILE="$REMEMBER_DIR/tmp/last-save.json"
 PID_FILE="$REMEMBER_DIR/tmp/save-session.pid"
-SESSION_DIR="$HOME/.claude/projects/$(session_dir_slug "$PROJECT")"
+SESSION_DIR="${CLAUDE_CONFIG_DIR:-$HOME/.claude}/projects/$(session_dir_slug "$PROJECT")"
 
 [ -f "$SAVE_SCRIPT" ] || exit 0
 
--- a/scripts/save-session.sh
+++ b/scripts/save-session.sh
@@ -99,7 +99,7 @@ for arg in "$@"; do
     esac
 done
 
-SESSION_DIR_PATH="$HOME/.claude/projects/$(session_dir_slug "$PROJECT_DIR")"
+SESSION_DIR_PATH="${CLAUDE_CONFIG_DIR:-$HOME/.claude}/projects/$(session_dir_slug "$PROJECT_DIR")"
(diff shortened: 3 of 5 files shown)


fdaviddpt:
Thanks for this — the diagnosis and the fix are both correct. Closing because it's already shipped, not because it's wrong.

`CLAUDE_CONFIG_DIR` support landed in #166, which merged before this PR was opened. This branch is based on `d9480ec`, so it couldn't have seen it. On current `main`:

- `pipeline/extract.py:53` reads `CLAUDE_CONFIG_DIR`
- `scripts/post-tool-hook.sh:45`, `scripts/save-session.sh:102` and `scripts/session-start-hook.sh:73` all resolve through a shared `claude_projects_dir()` helper (`detect-tools.sh:131`)
- `scripts/run-tests.sh:151` honours it too
- the "no session dir" warning in `post-tool-hook.sh:76` already interpolates `$(claude_projects_dir)` rather than a hardcoded path

The one substantive difference is the shape: this PR inlines `${CLAUDE_CONFIG_DIR:-$HOME/.claude}` at each of the three call sites, where `main` routes them through one helper. Given #177 and #158 are both open about rules that exist in more than one place and drift apart, we'd rather keep the single source of truth — so merging this would be a step backwards even though the behaviour is equivalent.

`main`'s `extract.py` also picked up two things this branch doesn't have: `rstrip("/\\\\")` so a trailing slash doesn't double up, and an explicit `HOME` lookup before `expanduser` so Windows test fixtures that patch only `HOME` still work.

Two observations from reviewing this **are** still live on `main`, and I'd rather not lose them:

1. `CLAUDE_CONFIG_DIR` still isn't path-normalised on Windows, unlike `PROJECT_DIR` in `resolve-paths.sh:111-125` — that's #169, still open.
2. `extract.py` doesn't `expanduser()` the variable while the sibling `home` lookup in the same function does, so a quoted `export CLAUDE_CONFIG_DIR="~/.claude-alt"` yields a literal tilde. Minor, but the asymmetry is real. I'll file it.

Sorry for the wasted effort — that's on the repo for not making the #166 overlap visible sooner.

Issue #142, opened 2026-07-24
NDC race: background subshell truncates now.md up to 180s after snapshotting it — saves landing in the window are silently lost


bonyohana:
## Mechanism

In `scripts/save-session.sh`, the NDC prompt snapshots `now.md` **synchronously** (line 231, `build-ndc-prompt`), then the compression runs in a **background subshell** with a 180s model-call timeout, and only afterwards truncates:

- line 231: snapshot into `$NDC_PROMPT`
- line ~237: `call-haiku ... 180` (in the backgrounded subshell)
- line 248: `: > "$MEMORY_FILE"`

So the truncation can land up to ~180s after the snapshot. The save cooldown defaults to **120s** (`scripts/save-session.sh:114`), i.e. a *different session's* save (or a `--force` save; the lock is released when the parent exits, and the parent exits without waiting on the subshell) can append a fresh entry to `now.md` inside the window. That entry was never in the snapshot, but `: >` erases it — never compressed into `today-*.md`, never seen again. No log line records the discrepancy.

## Suggested fix

Truncate-by-content, not blindly: have the subshell replace `now.md` only up to what it snapshotted — e.g. snapshot to a byte offset (`NDC_SRC_BYTES` is already computed at line ~229!) and, at completion, atomically rewrite `now.md` with whatever lies **beyond** that offset instead of `: >`:

```bash
tail -c +$((NDC_SRC_BYTES + 1)) "$MEMORY_FILE" > "$MEMORY_FILE.tmp" && mv "$MEMORY_FILE.tmp" "$MEMORY_FILE"
```

(mv-over is atomic on the same filesystem; worst case under a concurrent append is the pre-existing race, no worse.) That preserves any entry appended during the compression window at zero cost.

PR #173, opened 2026-07-26
fix(ndc): serialize now.md writers behind a lock (follow-up to #142)

Jutiphan:
## Credit where due

#142 fixed the headline bug: NDC compression used to `: > "$MEMORY_FILE"` —
a blind truncate — after handing `now.md` to a Haiku call that can take up
to 180s. By the time that landed, the parent had released the save lock
and exited, so a newer save could legitimately have appended an entry,
and the blind truncate erased it. #142's fix (keep only the bytes
actually compressed, drop the rest) is the right idea and stays in
place. This PR does not change that shape — it closes the gaps left by
*not holding a lock* around it.

## What's still broken

`save-session.sh`'s NDC step has three residual defects, all stemming
from the same root cause: nothing about it holds a lock.

**(a) Unsynchronized snapshot.** The byte-count read and the subsequent
`build-ndc-prompt` read are two separate, unlocked reads of `now.md`. A
save landing between them can change what's actually in the file, so
the byte count no longer describes what was fed to the summarizer.

**(b) Unsynchronized commit.** The closing `tail`/`mv` that commits the
partial truncate is itself unserialized against a concurrent append —
two NDC rounds, or an NDC round and a save's append, can interleave
arbitrarily.

**(c) The tail-failure branch still falls back to `: > "$MEMORY_FILE"`**
— the *original* #142 bug, preserved as an error path. If `tail` ever
fails (disk hiccup, permissions, etc.), the code path #142 specifically
added to avoid data loss regresses to exactly that data loss.

This is the shape of concurrency several worktree lanes plus two
accounts (`CLAUDE_CONFIG_DIR`) sharing one `REMEMBER_DIR` hits routinely.

## The fix

New `scripts/lib-now-lock.sh`, sourced by `save-session.sh`, giving
every `now.md` writer (save's append, NDC's snapshot, NDC's commit) a
single `flock`-guarded critical section:

- `now_locked <timeout_s> <command> [args...]` — runs a command inside
  a subshell holding `flock` on `${REMEMBER_DIR}/tmp/now.lock`. Returns
  the command's rc, or `99` on lock-acquire timeout.
- `now_append <memory_file> <content_file>` — the append, called via
  `now_locked` (replaces the unlocked `echo "" >> ...; cat ... >> ...`).
- `now_truncate_first <memory_file> <n_bytes>` — drops exactly the
  first `n` bytes via a sibling-temp-file + atomic `mv` (never an
  O_TRUNC window). If `n` exceeds the file's current size, or `tail`
  itself fails, `now.md` is left **completely untouched** and a nonzero
  rc is returned — never the blind `: >` of #142. The temp file is
  removed on every path.

`save-session.sh` changes:
- The append (Step 7) goes through `now_locked 30 now_append ...`.
- The byte-count snapshot and the `build-ndc-prompt` read are combined
  into one function (`_ndc_snapshot`) and run under one `now_locked 30`
  call, so they can never observe different states of `now.md` (closes
  (a)).
- The commit (today-*.md append + truncate) is combined into
  `_ndc_commit` and run under one `now_locked 60` call (closes (b)),
  and a truncate refusal is logged with its specific cause (lock
  timeout vs. bounds guard) instead of silently doing nothing (closes
  (c) — there is no more blind fallback to fall back to).

#164's day-attribution stamping (`NOW_DAY_FILE`, filing compressed
memory under the day it came from rather than the day the run happens
to fall on) is preserved unchanged — it now runs inside the same
locked commit it always logically belonged to.

## Test evidence

Existing `tests/test_ndc_truncate_race.py` (the #142 regression test)
is unaffected and still passes:
```
tests/test_ndc_truncate_race.py::TestNdcTruncateRace::test_entry_appended_during_compression_survives PASSED
tests/test_ndc_truncate_race.py::TestNdcTruncateRace::test_compressed_content_is_still_removed PASSED
2 passed in 5.05s
```
(Its shared harness in `test_save_session_gates.py` needed one line
added — `lib-now-lock.sh` to the list of scripts copied into the stub
plugin dir — since `save-session.sh` now sources it.)

New file: `tests/test_now_lock_concurrency.py`, exercising the library
directly:
- `test_two_concurrent_holders_never_overlap` — two real processes race
  on `now_locked`; asserts their critical sections are strictly
  serialized (start/end trace never interleaves).
- `test_refuses_when_n_exceeds_current_size` — the bounds guard.
- `test_tail_failure_leaves_file_untouched_not_emptied` — a fake
  failing `tail` on `PATH`; asserts `now.md` is byte-for-byte unchanged
  (defect (c)).

Before (unpatched `main`, HEAD `d9480ec` — `lib-now-lock.sh` does not
exist):
```
FAILED tests/test_now_lock_concurrency.py::TestNowLockedMutualExclusion::test_two_concurrent_holders_never_overlap
FAILED tests/test_now_lock_concurrency.py::TestNowTruncateFirstSafety::test_refuses_when_n_exceeds_current_size
FAILED tests/test_now_lock_concurrency.py::TestNowTruncateFirstSafety::test_tail_failure_leaves_file_untouched_not_emptied
3 failed in 0.15s
```

After (patched):
```
3 passed in 5.05s
```

### Regression suite

This machine has no `jq`, so `main`'s own `pytest tests/ -q` baseline
already fails a batch of tests unrelated to this change:

- Baseline (`main`, HEAD `d9480ec`): `38 failed, 625 passed, 58 skipped`
- Patched: `38 failed, 628 passed, 58 skipped` (identical pre-existing
  failures, +3 for this patch's new tests)

## Backward compatibility

None needed. `now.lock` is a new file under `${REMEMBER_DIR}/tmp/`,
created on first use; no existing file's format or location changes.

---
This PR is independent of the other four bug-fix PRs I'm submitting
alongside it and can be reviewed/merged in any order.

### DIFF
--- a/scripts/lib-now-lock.sh
+++ b/scripts/lib-now-lock.sh
@@ -1,0 +1,90 @@
+#!/bin/bash
+# ============================================================================
+# lib-now-lock.sh — serialized read/modify/write access to now.md
+# ============================================================================
+#
+# DESCRIPTION
+#   Follow-up to #142. That fix replaced the blind `: > "$MEMORY_FILE"` NDC
+#   truncate with a partial truncate (drop only the bytes that were actually
+#   compressed, keep anything a concurrent save appended after the snapshot)
+#   — the right idea, but every step around it is still unsynchronized:
+#
+#     1. The byte-count snapshot (`wc -c`) and the `build-ndc-prompt` read
+#        that follows it are two separate, unlocked reads of now.md. A save
+#        landing in between makes the snapshot disagree with what was
+#        actually fed to the summarizer — the eventual truncate then drops
+#        bytes that were never summarized.
+#     2. The closing `tail`/`mv` that commits the truncate is itself
+#        unserialized against a concurrent append landing at the same time.
+#     3. The `tail`-failure branch still falls back to `: > "$MEMORY_FILE"`
+#        — the original #142 bug, preserved as an error path.
+#
+#   This library gives every now.md writer (save's append, NDC's snapshot,
+#   NDC's commit) a single flock-guarded critical section, so none of the
+#   three windows above can open.
+#
+# USAGE
+#   source "$(dirname "$0")/lib-now-lock.sh"
+#   now_locked 30 now_append "$MEMORY_FILE" "$HAIKU_TEXT_FILE"
+#
+# EXIT CODES (now_locked)
+#   rc of the wrapped command, or 99 if the lock could not be acquired
+#   within the given timeout.
+#
+# ============================================================================
+
+[ -n "${_LIB_NOW_LOCK_LOADED:-}" ] && return 0
+_LIB_NOW_LOCK_LOADED=1
+
+# now_locked <timeout_s> <command> [args...]
+# Runs the command (function or external) inside a subshell holding an
+# flock on now.lock, so it cannot interleave with any other now_locked
+# caller. Returns the command's exit code, or 99 on lock-acquire timeout.
+now_locked() {
+    local _timeout="$1"; shift
+    (
+        flock -w "$_timeout" 9 || exit 99
+        "$@"
+    ) 9>>"${REMEMBER_DIR}/tmp/now.lock"
+}
+
+# now_append <memory_file> <content_file>
+# Blank-line separator + content, same shape as the unlocked append it
+# replaces. Call only via now_locked.
+now_append() {
+    local mem="$1" content="$2"
+    echo "" >> "$mem" && cat "$content" >> "$mem"
+}
+
+# now_truncate_first <memory_file> <n_bytes>
+# Drops ONLY the first n bytes (the span already compressed into
+# today-*.md), preserving anything appended after the snapshot. Call only
+# via now_locked, immediately after taking the same snapshot count under
+# the same lock (see save-session.sh's NDC step).
+#
+# Writes the tail to a sibling temp file and `mv`s it over $mem (same
+# directory, same filesystem — an atomic rename, not an O_TRUNC window).
+# If `n` exceeds the file's current size, or `tail` itself fails, $mem is
+# left completely untouched and a nonzero rc is returned — never the
+# blind `: > $mem` of #142. The temp file is removed on every path.
+now_truncate_first() {
+    local mem="$1" n="$2" tmp size rc
+    [ -z "$n" ] && return 0
+    [ "$n" -eq 0 ] && return 0
+
+    size=$(wc -c < "$mem") || return 1
+    if [ "$n" -gt "$size" ]; then
+        return 1
+    fi
+
+    tmp="${mem}.tail.$$"
+    if tail -c +"$((n + 1))" "$mem" > "$tmp"; then
+        mv -f "$tmp" "$mem"
+        rc=$?
+        [ "$rc" -ne 0 ] && rm -f "$tmp"
+    else
+        rc=$?
+        rm -f "$tmp"
+    fi
+    return $rc
+}
--- a/scripts/save-session.sh
+++ b/scripts/save-session.sh
@@ -58,6 +58,7 @@ source "$(dirname "$0")/resolve-paths.sh"
 source "$(dirname "$0")/detect-tools.sh"
 source "$(dirname "$0")/bootstrap-dirs.sh"
 source "$(dirname "$0")/log.sh"
+source "$(dirname "$0")/lib-now-lock.sh"
 log "hook" "save-session: PROJECT_DIR=$PROJECT_DIR PIPELINE_DIR=$PIPELINE_DIR PYTHON=$PYTHON REMEMBER_DIR=$REMEMBER_DIR"
 
 LOCK_FILE="${REMEMBER_DIR}/tmp/save.lock"
@@ -350,8 +351,8 @@ fi
 if [ ! -s "$MEMORY_FILE" ]; then
     printf '%s\n' "$TODAY_DATE" > "$NOW_DAY_FILE" 2>/dev/null || true
 fi
-echo "" >> "$MEMORY_FILE" 2>/dev/null || { log "write" "ERROR: cannot write now.md"; exit 1; }
-cat "$HAIKU_TEXT_FILE" >> "$MEMORY_FILE"
+now_locked 30 now_append "$MEMORY_FILE" "$HAIKU_TEXT_FILE" \
+    || { log "write" "ERROR: cannot write now.md (lock timeout or write failed)"; exit 1; }
 log "write" "appended: $(head -1 "$HAIKU_TEXT_FILE" | cut -c1-80)"
 cd "$PIPELINE_DIR" && $PYTHON -m pipeline.shell save-position "$LAST_SAVE_FILE" "$SESSION_ID" "$POSITION"
 log "write" "position → $POSITION"
@@ -393,12 +394,22 @@ TODAY_FILE="${REMEMBER_DIR}/today-${NDC_DAY}.md"
 if [ "$RUN_NDC" = true ]; then
     log "ndc" "now.md → today-${NDC_DAY}.md"
     date +%s > "$NDC_MARKER"
-    NDC_SRC_BYTES=$(wc -c < "$MEMORY_FILE" | tr -d ' ')
     NDC_PROMPT=$(mktemp "${TMPDIR:-/tmp}"/remember-ndc-XXXXXX)
 
-    cd "$PIPELINE_DIR" && $PYTHON -m pipeline.shell build-ndc-prompt "$MEMORY_FILE" "$NDC_PROMPT"
+    # Snapshot the byte-count and the build-ndc-prompt read together, under
+    # now.lock — so the count can never disagree with what was actually fed
+    # to the summarizer (a save landing between an unlocked `wc -c` and the
+    # prompt build was defect (a) of #142's incomplete follow-up).
+    _ndc_snapshot() {
+        wc -c < "$MEMORY_FILE" | tr -d ' '
+        (cd "$PIPELINE_DIR" && $PYTHON -m pipeline.shell build-ndc-prompt "$MEMORY_FILE" "$NDC_PROMPT") >&2
+    }
+    NDC_SRC_BYTES=$(now_locked 30 _ndc_snapshot) || NDC_SRC_BYTES=""
+    if [ -z "$NDC_SRC_BYTES" ]; then
+        log "ndc" "ERROR: snapshot lock timeout — skipping NDC this round"
+    fi
 
-    if [ -s "$NDC_PROMPT" ]; then
+    if [ -n "$NDC_SRC_BYTES" ] && [ -s "$NDC_PROMPT" ]; then
         (set +e  # don't inherit set -e — a haiku non-zero exit must not kill the subshell
             NDC_ERR=$(mktemp "${TMPDIR:-/tmp}"/remember-ndc-err-XXXXXX)
             # 180s (not the 120s default): NDC compresses a whole now.md.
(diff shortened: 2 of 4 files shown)


fdaviddpt:
Thanks for this, and sorry for the slow reply. The diagnosis is right and the design is sound — the problem is a dependency that does not exist on the maintainer's platform.

## Blocker: `flock` is not available on macOS

`scripts/lib-now-lock.sh:46`:

```sh
flock -w "$_timeout" 9 || exit 99
```

No `command -v flock` guard, no fallback. Verified on a stock Mac just now:

```
$ command -v flock        -> (nothing)
$ ls -la /usr/bin/flock   -> No such file or directory
```

`.github/workflows/tests.yml` includes `macos-latest` with no `brew install util-linux` step, and the GitHub macOS runner images do not ship it either.

Measured on a stock Mac, base `d9480ec` vs your `ad41354`, same machine:

- **23 passed → 16 passed, 7 newly failed**
- among them **`test_ndc_truncate_race.py`** — the #142 regression test this PR's description states is "unaffected and still passes". It fails with `returncode == 1` instead of `0`.
- `now_locked 5 echo hi` → `flock: command not found`, returns 99, which the caller reports as "lock timeout"
- simulating the real append path: `now_locked 30 now_append ...` fails, hits `|| { log ERROR; exit 1; }`, **the append never happens and the script aborts**

Net effect on macOS: a rare data race becomes a 100%-reproducible failure of every save, plus NDC compression permanently disabled (`NDC_SRC_BYTES` empty → "skipping NDC this round", every round). Your own `TestNowLockedMutualExclusion::test_two_concurrent_holders_never_overlap` correctly fails here too — the test is good, the environment it was validated in just had `flock`. I read the "After (patched): 3 passed in 5.05s" as Linux-only validation.

## The fallback is trickier than it looks — a warning before you spend time on it

My first instinct was `python3 -c` with `fcntl.flock` on the inherited fd, since python3 is already a hard dependency of every hook. **That does not work.** An advisory lock belongs to the open file description, so it is released the moment the Python child exits — the shell would carry on believing it holds a lock it does not. Silent, and worse than the current bug.

The lock has to be held by something that lives as long as the critical section. Options that actually work in bash:

- **`mkdir`-based lock.** `mkdir` is atomic — it either creates or fails, no check-then-act — and is available everywhere. Spin with a sleep until the timeout. Simplest thing that is correct.
- **`flock` when present, `mkdir` otherwise**, if you want to keep the fast path on Linux.

Relevant: we have just opened #182 to move `save.lock` to `mkdir` acquisition with `mv`-based stale takeover, for a closely related multiple-holders bug. It would be worth these two locks sharing one primitive rather than each growing its own — happy to coordinate so this does not end up as a third implementation of the same rule (see #177).

## What is already right

None of the above is about the design:

- an flock'd FD rather than a pidfile means the kernel releases on process death — no stale-lock class of bug, notably not the one #168 just fixed for `save.lock`
- `now_truncate_first`'s bounds guard plus sibling-temp-file and atomic `mv` is correct; both bounds tests go red against base
- keeping the 180s Haiku call outside the lock, with the byte-offset truncate preserving concurrently-appended bytes, is the right shape
- all 3 new tests in `test_now_lock_concurrency.py` go red against base — no tests that pass either way

## Minor

- rc 99 covers both "lock timeout" and "flock missing", with the same log line, so the real cause is undiagnosable from logs
- on genuine >30s contention the losing append is dropped rather than retried, so that save's entry is lost. Low probability given sub-millisecond appends, but it is a new failure mode the racy-but-always-succeeds code did not have.

Holding this open rather than closing it — the work is good and the remaining piece is the portability layer. Tell me if you would rather I take that on and land it with your commit intact.

fdaviddpt:
Update so you are not waiting on silence: #182 is the plan, and it is written up as a full design brief — one `mkdir`-based primitive in `scripts/lib-lock.sh`, used by both `save.lock` and your `now.lock`, so the repo does not end up with three lock implementations (#177).

Your work is being kept, explicitly. The brief says so: `now_truncate_first`'s bounds guard, the sibling-temp-file plus atomic `mv`, and keeping the 180s Haiku call outside the lock with a byte-offset truncate that preserves concurrent appends — that is the correct shape and it stays. Only `flock -w 9` gets swapped for `lock_acquire`. Your three tests in `test_now_lock_concurrency.py` all go red against base, which is what a real regression test looks like.

Two practical notes. The branch now conflicts with `main` — #178 and #168 both landed in `save-session.sh` since you opened it. And the trap worth carrying into any rework: N=2 concurrency tests pass on broken locking (0/40 multi-winner at N=2, 11/40 at N=8), which is exactly how the previous fix shipped looking correct.

Tell me which you prefer: you rebase and take the portability layer from #182, or I land it with your commit intact. Either is fine — I would rather you choose than have me decide by acting.

fdaviddpt:
Thanks for this — the analysis is good and defect (c) is real. I verified it before writing this rather than taking the description on trust, and I want to be precise about which parts I confirmed, because the conclusion I reach is different from the patch.

## (c) is real, and it is the worst of the three

`scripts/save-session.sh:541`, the `else` arm of the `tail -c` branch:

```sh
else
    rm -f "$NDC_TAIL"
    : > "$MEMORY_FILE"
    rm -f "$NOW_DAY_FILE"
fi
```

That is the #142 blind truncate, alive as an error path. The comment ten lines above it explains why #142 removed the truncate from the success path — and the failure path, where the file's contents are *least* likely to be recoverable, still does the thing #142 was about. You are right that this regresses to the original data loss, and the fix is small and independent of the locking question. It should not wait on the rest.

(a) and (b) I also read as real: the byte-count read and the `build-ndc-prompt` read are separate unlocked reads, and the commit is unserialized against a concurrent append. The failure needs a narrow interleaving, but the multi-worktree + two-`CLAUDE_CONFIG_DIR` setup you describe is exactly this repo's normal usage, not an exotic one.

## Why I can't take the patch as written

The critical section is built on `flock`, and this repo rejected `flock` deliberately. `scripts/lib-lock.sh:27-32`:

```
# Not `flock`: absent on macOS (no /usr/bin/flock, and the GitHub macos runner
# image does not ship util-linux). Not `python3 -c` with `fcntl.flock` either —
# an advisory lock belongs to the open file description and is released the
# moment the Python child exits, so the shell would carry on believing it holds
# a lock it does not. Silent, and worse than no lock at all.
```

`which flock` on the maintainer's machine returns nothing. So on macOS `now_locked` cannot acquire, and every `now.md` writer — including the ordinary save append in Step 7, which is on the hot path of every session — ends up either failing or running unlocked. A fix for a rare interleaving that costs saves on the primary development platform is a bad trade, and it is the same shape as a change we turned down recently: a patch that was correct in the reporter's environment and a silent total outage in ours.

The pre-emptive rejection of the `python3 fcntl.flock` workaround in that same comment is worth reading too — it closes off the obvious portability patch on top.

## What I'm doing instead

- Taking (c) as its own change now: the `else` arm must leave `now.md` untouched and return nonzero, never truncate. Your `now_truncate_first` has the right post-condition ("if `tail` fails, `now.md` is left completely untouched") and I'm keeping that contract.
- Building the critical section for (a)/(b) on the existing `lib-lock.sh` — the `mv`-based single-winner takeover already in this repo, which was written for this exact portability constraint — rather than introducing a second, less portable locking primitive beside it.

Not merging this as-is, but the bug report earned the fix. I'll credit the PR in the changelog entry. If you want to rework it onto `lib-lock.sh` yourself, say so and I'll leave it to you rather than duplicating your work.

fdaviddpt:
Status update, with the accounting rather than a "still open" note.

I went back through `save-session.sh` on `main` today and checked each of your three defects against the code as it now stands:

- **(a) unsynchronized snapshot — closed.** Both the `wc -c` and the `build-ndc-prompt` read happen in the parent, before the subshell is backgrounded, so they run under the lock acquired at line 102.
- **(c) the `: > "$MEMORY_FILE"` failure path — closed.** This was the one you called the worst of the three and I agreed. now.md is now left completely untouched when `tail` fails, and the resulting trade (a span that may be summarized twice) is documented in place, including why rolling the `TODAY_FILE` append back was rejected as the same bug moved one file over.
- **(b) unsynchronized commit — still open.** You were right and it is unmitigated. The `tail` → `mv` pair runs inside the subshell backgrounded at line 571, after the parent's EXIT trap has released the lock. An append landing between the read and the `mv` is erased, offset already advanced, nothing logged.

I have filed (b) as #223, with your diagnosis credited and the note that #182's `mkdir`-based `lock_acquire` now makes the fix available without the `flock` dependency that blocked this PR on macOS. So the thing that stopped your patch is what produced the primitive that unblocks it.

Leaving this PR open — the write-up in it is still the best description of this area, and closing it is your call, not mine. If you would rather rework it onto `lib-lock.sh` now that the primitive exists, say so and it is yours; if you would rather hand it off, that is fine too and the credit stays either way.

Sorry it sat. That is on me, not on the patch.

fdaviddpt:
Thank you for this, and sorry it sat — you've been waiting since the 26th. Here is a straight assessment of what's left of it, because the ground moved twice underneath it since you opened it.

## Your three windows, against `main` today

**Window 2 — the unserialized `tail`/`mv` commit: closed**, by #224 (`32a3dae`), which re-acquires the save lock around the whole commit and re-reads `now.md`'s size under it. That landed on the 29th, three days after you opened this.

**Window 3 — "the `tail`-failure branch still falls back to `: > "$MEMORY_FILE"`": not present.** I checked: the only `> "$MEMORY_FILE"` occurrences on `main` are an *append* (`cat … >>`, line 435) and a comment. This claim was almost certainly true when you wrote it and is stale now, not wrong-at-the-time.

**Window 1 — the `wc -c` snapshot and the `build-ndc-prompt` read being two separate unlocked reads: still open.** This is the real remaining substance of your PR.

One correction on its harm, and it matters for how a fix should be shaped. The order in `save-session.sh` is `NDC_SRC_BYTES=$(wc -c …)` at line 477, *then* `build-ndc-prompt` at line 480. So an append landing between them makes the summarized span **larger** than the snapshot, and truncating the snapshot count leaves already-summarized bytes in `now.md` — that's **duplication**, not the erasure your description implies. Still a real defect, opposite direction.

## What changed today

`#234` just merged: `today-*.md` now has its own lock (`scripts/lib-staging-lock.sh`), because the NDC append and consolidation's retire loop held *different* locks over the same file, and an append landing among the retire loop's `wc -c`/`head`/`tail`/`mv` was sealed into a file nothing globs again. Reproduced against the shipped scripts with a second process before fixing — the entry existed only inside the retired `.done.md`.

That is a different window from yours, but it settles the question your PR raises for the whole area, which is **lock ordering**. The property #234 buys deliberately: *nothing holds the staging lock while acquiring another, and nothing acquires it while holding `save.lock`.* There is no order to get wrong.

**So your PR is not blocked by it, and the opposite-order deadlock I was worried about is not reachable** — different files, and by construction no path holds both.

## The one thing that does need changing before this can land

**`now_locked` uses `flock`, and this repo deliberately avoids it.** `flock` is absent on macOS and on the GitHub `macos-latest` runner; `scripts/lib-lock.sh` exists for exactly that reason and documents it. So the mechanism needs porting onto the existing primitive — your *analysis* is sound and your critical-section boundaries look right to me, but as written the library would be a no-op or an error on half the supported platforms.

Two smaller notes if you take it further:

- #234 added `lib-staging-lock.sh` as a worked example of the shape: a lock that owns one resource, sourced on top of `lib-lock.sh`, with its rationale in the header. `now.md` is a legitimate second instance of that pattern.
- Your PR substitutes a `now.md` lock for the save-lock re-acquisition that #224 added around the NDC commit. That is still compatible — #234's serialization runs through the staging lock around the *append*, which your PR does not touch — but the interaction is worth stating explicitly in the PR body so the next reader doesn't have to derive it.

## Credit where it's due

Your framing of the three windows is what made this tractable: the first one is still open, and it is open *because* you separated it from the commit-side problem that #142 and #223 had already absorbed all the attention. The `#142`-recognition — that the partial truncate was the right idea and the gap was the absence of a lock around it — is the same conclusion #224 reached independently three days later.

If you'd rather not do the `flock` port, say so and it can be picked up from here with your analysis credited — no hard feelings either way, and either route is genuinely useful.

fdaviddpt:
One factual update since the assessment above, because it changes what a rebase of this branch would look like — everything else in that comment still stands, including that **Window 1 is the real remaining substance** and that the `flock` → `lib-lock.sh` port is the blocker.

**The append your `now_append` replaces no longer exists.** That comment cites it at `save-session.sh:435` as `echo "" >> …; cat … >>`. #247/#250 landed tonight (`70926c3`) and replaced it with a staged copy renamed over `now.md` — `save-session.sh:477-496`.

The reasoning is adjacent to yours rather than a rejection of it: a lock excludes other **writers**, and the reader that mattered structurally cannot take one. `session-start-hook.sh` sources five libraries, `lib-lock.sh` is not among them, so `lock_acquire` is undefined in that process, and its memory injection is a bare `cat`. A reader lock was considered and rejected — #227 measured that hook at a p50 of 8.7s, #230 exists to cut its per-tool-call prefix, and #204 is an external report of the plugin blocking a user's prompts. Putting session start behind a lock held across a summarize call is a worse trade than the tear it would close.

So the two mechanisms divide cleanly, and it is worth stating because it affects how a re-scoped PR should be framed:

- **A rename protects readers that cannot participate in the locking at all.** That is now handled for the append path.
- **A lock is still the only answer where two operations must see the same bytes** — which is exactly your Window 1, where `wc -c` and `build-ndc-prompt` are separate unlocked reads. A rename does nothing there.

Net effect on this branch: `now_append` is now a rewrite of code that is gone and will conflict, while your `now_locked` primitive and the snapshot critical section — the parts the assessment above called sound — are untouched by tonight's merge. If you take it further, a PR aimed only at Window 1 would be substantially smaller than this one.

Still your call, and the offer above stands unchanged: if you'd rather not do the `flock` port, it can be picked up from here with your analysis credited.

fdaviddpt:
Closing this, and I want to be precise about why, because the reason is "you were right and it has all been fixed" rather than "this was wrong".

I re-checked your three defects against `main` today (`70926c3`), reading the code rather than the history:

- **(a) unsynchronized snapshot — closed.** `save-session.sh:539` (`wc -c`) and `:542` (`build-ndc-prompt`) both run in the parent, which holds `save.lock` from `:108` until its EXIT trap; the subshell only starts at `:545`. I confirmed the writer set is closed — the only writers of `now.md` anywhere in the repo are Step 7's `mv` (`:496`) and the NDC commit's `mv` (`:723`), both under that lock. Everything else is a reader.
- **(b) unsynchronized commit — closed by #224.** The `wc -c`, bounds guard, `tail` and `mv` all sit inside `lock_acquire "$LOCK_DIR" "$NDC_COMMIT_LOCK_TIMEOUT"` at `:647`.
- **(c) the blind-truncate failure path — closed.** No such truncate exists on `main`. The `else` arm at `:759` leaves `now.md` untouched and logs, and the duplication trade is documented in place.

Your framing of them as three separate windows is what made them tractable one at a time, and #224 reached your (b) conclusion independently three days after you wrote it. `now_truncate_first`'s bounds guard, the sibling temp file and the atomic `mv` are the same shape `main` converged on. That is a good report.

On the branch itself: `git merge-tree` gives four conflict hunks in `save-session.sh`, and the two substantive ones are total rewrites on both sides. `now_append` replaces code that no longer exists — #247/#250 turned Step 7 into a staged copy renamed over `now.md`, because a writer-side lock cannot reach `session-start-hook.sh`'s bare `cat`, which never sources `lib-lock.sh`. So there is nothing left to rebase onto.

One thing worth carrying forward regardless, because it is the failure mode this repo is most prone to: `flock -w "$_timeout" 9 || exit 99` maps `command not found` (rc 127) onto the same 99 as a genuine timeout. On macOS — where `flock` is absent, verified again just now — every caller would log "lock timeout" forever while the real cause was a missing binary. An absence produced by the environment read as an absence in the world. Worth guarding in any future lock wrapper, ours included.

Credit stands in the changelog. Thanks for the care in the write-up — it was the best description of this area for several weeks, and it is why the fixes landed in the order they did.

Issue #131, opened 2026-07-17
_child_env strips CLAUDE_CODE_OAUTH_TOKEN along with session vars, breaking nested claude -p auth for setup-token users


socialmenteagency:
## Summary

`_child_env()` in `pipeline/haiku.py` strips every environment variable whose name starts with `CLAUDE_CODE_`, to keep session-identifying vars (e.g. `CLAUDE_CODE_SESSION_ID`) from leaking into the nested `claude -p` and making it look like a resumable child session (#95). `CLAUDE_CODE_OAUTH_TOKEN` — the long-lived credential produced by `claude setup-token`, meant for exactly this kind of headless/background CLI call — matches that same prefix by naming coincidence, so it gets stripped along with everything else. The nested `claude -p` ends up with no credentials at all, even when the user has correctly set `CLAUDE_CODE_OAUTH_TOKEN` and a bare `claude -p` call in the same shell works fine.

This hit hardest in an Agent-SDK-hosted session (not a plain terminal install): the interactive session authenticates through the host, not through a file/keychain credential the standalone CLI can reuse, so `claude setup-token` + the env var was the correct fix for the standalone CLI — but `remember`'s own stripping undid it for its own subprocess calls specifically.

## Environment

- remember 0.8.3 (bug still present on `main`, `pipeline/haiku.py` `_child_env()`, current lines ~96-113)
- Claude Code CLI 2.1.202, macOS (Darwin 25.5.0)
- Auth method: `claude setup-token` → `CLAUDE_CODE_OAUTH_TOKEN`

## Reproduction

1. Set `CLAUDE_CODE_OAUTH_TOKEN` in the shell (via `claude setup-token`), confirm it works standalone:
   ```
   claude auth status
   # {"loggedIn": true, "authMethod": "oauth_token", ...}
   echo "hi" | claude -p --output-format json --model haiku
   # succeeds
   ```
2. Run the plugin's own pipeline with the same env:
   ```
   bash scripts/save-session.sh --force
   ```
3. It still fails with the (also-tracked-in-#129) undiagnosable message:
   ```
   [haiku] ERROR: call-haiku error: claude exited 1: 
   ```
   Applying the #129 stderr/stdout patch surfaces the real reason: `Not logged in · Please run /login` — despite the parent shell being correctly authenticated via `CLAUDE_CODE_OAUTH_TOKEN`.

## Suggested fix (tested)

[...]

Issue #129, opened 2026-07-16
call_haiku loses the real error: claude -p --output-format json reports failures on stdout, not stderr


bigtopmultimedia:
## Summary

When the `claude -p` subprocess fails, `call_haiku` raises `RuntimeError(f"claude exited {result.returncode}: {stderr}")` — but with `--output-format json` the CLI reports failures as an **error JSON on stdout** (`"is_error": true`) and leaves **stderr empty**. The result is that every subprocess failure logs as the undiagnosable:

```
[haiku] ERROR: call-haiku error: claude exited 1:
```

In my case this hid a CLI auth outage for ~7 weeks: the standalone CLI had been logged out (interactive sessions kept working because the desktop app injects its own auth), so every auto-save and consolidation died with the empty message above, `now.md` went stale, and daily files were never compressed into `recent.md`/`archive.md`.

## Environment

- remember 0.8.3 (bug still present on `main`, `pipeline/haiku.py` lines 176–178)
- Claude Code CLI 2.1.197, macOS (Darwin 25.5.0)

## Reproduction

Log the CLI out (or run on any machine where `claude auth status` reports `"loggedIn": false`), then run the pipeline:

```
cd <plugin-dir> && python3 -m pipeline.shell consolidate <staging_dir> <recent.md> <archive.md>
```

→ `RuntimeError: claude exited 1: ` (empty detail).

Running the same `claude -p --output-format json ...` command manually shows what actually happened — exit code 1, empty stderr, and on **stdout**:

```json
{"type":"result","subtype":"success","is_error":true,"result":"Not logged in · Please run /login", ...}
```

## Suggested fix (tested)

In `call_haiku`'s non-zero-exit branch, fall back to the stdout JSON's `result` field when stderr is empty (then raw stdout, capped):

```python
    if result.returncode != 0:
        detail = result.stderr.strip()
        if not detail:
            # With --output-format json the CLI reports failures as an error
            # JSON on STDOUT (is_error: true) and leaves stderr empty — e.g.
            # a logged-out CLI prints {"result": "Not logged in · Please run
            # /login"} and exits 1. Surface that message, else every failure
            # logs as the undiagnosable "claude exited 1: ".
            try:
                payload = json.loads(result.stdout)
            except json.JSONDecodeError:
                payload = None
            if isinstance(payload, dict):
                detail = str(payload.get("result") or "").strip()
            if not detail:
                detail = result.stdout.strip()[:500]
[...]

PR #179, opened 2026-07-26
fix(haiku): read persisted OAuth token from the Windows registry when the host redacts it

cemkizildag1907:
# fix(haiku): read persisted OAuth token from the Windows registry when the host redacts it

## Problem

On Windows, under the Claude Code **desktop / Agent SDK host**, the background
pipeline saves nothing: every `call_haiku` fails and the failure is now visible
thanks to #129 as:

```
Failed to authenticate: OAuth session expired and could not be refreshed
```

This happens even after the operator has run `claude setup-token` and made the
token persistent with `setx CLAUDE_CODE_OAUTH_TOKEN "<token>"`.

## Root cause

`_CHILD_ENV_KEEP` (#131) preserves `CLAUDE_CODE_OAUTH_TOKEN` across the
parent-session strip — but only if it is in the environment to begin with. The
Claude Code desktop / Agent SDK host on Windows **redacts
`CLAUDE_CODE_OAUTH_TOKEN` from every spawned tool and hook subprocess** (a
deliberate boundary: it must not leak the OAuth credential to arbitrary
children). So `os.environ` holds nothing for `_child_env` to keep, and the
nested `claude -p` is unauthenticated again — the same silent-save outage as
#129, on a machine where auth was correctly set up.

Verified on this environment:

| Probe | Result |
|---|---|
| `setx` → `HKCU\Environment` | token persisted (len 108) |
| Token in a Claude-Code-spawned subprocess env (after full reboot) | **absent** |
| `winreg` read of `HKCU\Environment` from that subprocess | **succeeds** |
| `call_haiku` after injecting the registry value | succeeds |

A reboot does **not** fix it — it is not an OS env-propagation gap but a host
boundary. But the token the user stored with `setx` is still readable from the
per-user registry.

## Fix

`call_haiku` gains a Windows-only, best-effort fallback: when
`CLAUDE_CODE_OAUTH_TOKEN` is absent from the child env, read it from
`HKCU\Environment` (where `setx` writes) and inject it, so the nested CLI can
authenticate.

```python
def _inject_persisted_oauth_token(env: dict[str, str]) -> dict[str, str]:
    if os.name != "nt" or env.get("CLAUDE_CODE_OAUTH_TOKEN"):
        return env
    try:
        import winreg
        with winreg.OpenKey(winreg.HKEY_CURRENT_USER, "Environment") as key:
            token, _ = winreg.QueryValueEx(key, "CLAUDE_CODE_OAUTH_TOKEN")
    except OSError:
        return env
    if isinstance(token, str) and token.strip():
        env["CLAUDE_CODE_OAUTH_TOKEN"] = token
    return env
```

Called right after `env = _child_env()`.

**Safety / scope:**
- Windows-only (`os.name == "nt"`); a no-op everywhere else.
- Never overrides a value already in the environment (env wins; registry is a last resort).
- Read-only registry access; any failure leaves `env` untouched — never raises.
- No new config surface, no new dependency (`winreg` is stdlib on Windows).

## Testing

`python -m pytest tests/test_haiku.py` → **70 passed** (3 new):

- `test_call_haiku_reads_oauth_token_from_registry_when_host_redacts`
- `test_call_haiku_registry_fallback_does_not_override_env_token`
- `test_call_haiku_registry_fallback_is_windows_only`

The tests inject a fake `winreg` into `sys.modules` so they run on Linux/macOS CI.

## Related

Follows #129 (surface the JSON auth error) and #131 (keep `CLAUDE_CODE_OAUTH_TOKEN`).
This covers the case those two don't: the token being absent from the child env entirely.


fdaviddpt:
Thank you for this — the diagnosis is excellent. The probe table (persisted via `setx`, absent from the spawned subprocess, readable via `winreg`, works once injected) isolates the failure precisely, and the code is careful: Windows-only, never overrides an existing value, and the tests assert `OpenKey.assert_not_called()` on both negative paths. That's better evidence than most reports arrive with.

I'm not going to merge it in this form, and the reason isn't code quality.

Your own analysis names the mechanism: the host "redacts `CLAUDE_CODE_OAUTH_TOKEN` from every spawned tool and hook subprocess (a deliberate boundary: it must not leak the OAuth credential to arbitrary children)". This change exists to defeat that boundary. We'd be shipping code that reaches into per-user OS credential storage to recover a secret the platform intentionally withheld from hook subprocesses — and this plugin *is* an arbitrary child, spawned in the background on every save.

Even granting that the token only ever reaches a nested `claude -p` and never leaves the machine (I checked — no network, no logging of the value, no second sink), it sets a precedent I don't want in this codebase: "read the credential from OS storage when the host won't hand it to us." That's a pattern which is easy to extend and hard to argue against later.

## What I'd take instead

The problem is real and your users are stuck. The difference I care about is **consent**: the operator explicitly giving this plugin a token, rather than the plugin recovering one that was withheld from it.

Concretely — a token the plugin reads from its own configuration:

- `REMEMBER_OAUTH_TOKEN` env var, and/or a `haiku.oauth_token` key in `~/.remember/config.json`
- `call_haiku` uses it only when `CLAUDE_CODE_OAUTH_TOKEN` is absent from the child env — same precedence logic you already wrote
- documented as "the plugin will hold this credential and pass it to a nested CLI", so setting it is a deliberate act

Same outcome on your machine, no boundary circumvented, and it works on every platform rather than Windows-only. Most of your diff survives — `_inject_persisted_oauth_token` keeps its shape and its tests, only the source changes.

If you'd rather not rework it, say so and I'll open a follow-up issue crediting the diagnosis to you; the analysis is the hard part and it shouldn't be lost either way.

## Two code notes, if this comes back in any form

1. `_inject_persisted_oauth_token` is documented as "never raises", but `import winreg` is inside the `try` and only `OSError` is caught. `ImportError` is not an `OSError`, so an import failure propagates out of `call_haiku`. Narrow in practice, but the docstring promises otherwise.
2. Nothing validates the recovered value — any string becomes a credential, so a stale token produces the same auth failure with a more confusing origin. Worth at least a shape check before injecting.

Also worth stating plainly: the suite skips Windows, so the mocks prove the plumbing and nothing more. Your probe table is the only real verification that exists, and we can't reproduce it here. That's not a knock on your work — it's a gap in ours.

cemkizildag1907:
Thank you — the consent distinction is the right call, and I've reworked the PR to it. You're also right that it's the better design on the merits: it's now cross-platform rather than Windows-only, and it doesn't touch OS credential storage at all.

**What changed (force-pushed):**

- Dropped the `winreg` path entirely. The fallback token now comes from operator config: `REMEMBER_OAUTH_TOKEN` env var, or `haiku.oauth_token` in `config.json` (`$REMEMBER_DIR/config.json`, else `~/.remember/config.json`).
- Same precedence you called out: used **only** when the child env has no `CLAUDE_CODE_OAUTH_TOKEN`; a host-provided token always wins.
- The helper kept its shape (`_inject_configured_oauth_token`), only the source changed — so most of the diff and the test structure survive, as you predicted.

**Both code notes addressed:**

1. *"never raises"* — there's no `import winreg` anymore, so the `ImportError`-escapes-`OSError` gap is gone. The config read is wrapped in `try/except (OSError, ValueError)` and the env read can't raise, so the docstring is now honest.
2. *validation* — `_looks_like_token` shape-checks before injecting (non-empty, whitespace-free, length ≥ 20). A blank or malformed entry is ignored, so it surfaces at configuration time rather than as a confusing 401 from a garbage token. There's a test for exactly this.

**Docs:** README (config-key table + env-var table), `config.example.json` (`haiku.oauth_token` + a `_comments` note framing it as "the plugin holds this credential and passes it to a nested CLI — set it deliberately"), and a CHANGELOG entry.

**Tests:** configured env token used when the child env lacks one; host token not overridden; `config.json` fallback; malformed value rejected. The full `test_haiku.py` suite is green. Your point about the suite skipping Windows still stands — these mocks prove the resolution/precedence logic, not the host behaviour itself; the original probe table remains the only real-host verification.

fdaviddpt:
Merged as `0ffa30f`. Thank you for reworking it — you turned a rejection into a better design in under an hour, and the result is cross-platform where the original was Windows-only.

Being straight with you about what happened after the merge, since it touches your code: I ran a review pass on the merged change and it found two things, both now fixed in #184.

The first is the one that matters, and it is the goal your own PR set: a malformed token was refused **silently**. `_looks_like_token` returned `False`, the value was dropped, and `call_haiku` carried on unauthenticated — producing exactly the confusing 401 the fallback exists to prevent, now with a more confusing origin, since the operator cannot tell "rejected" from "never set". The docs we both wrote said it "fails at configuration time". Nothing failed. A refusal is now logged with its source and length, never the value, and into the daily log rather than stderr — `save-session.sh` captures `call-haiku`'s stderr and only echoes it when the call *fails*, so a warning there would have vanished on the path that matters.

The second: `_configured_oauth_token` hand-rolled its own config layer order instead of reading the merged `REMEMBER_CONFIG` that `lib-memory-dir.sh` already exports. My review comment asked for the config key without pointing you at the existing resolver, so that one is on me.

Neither changes the shape you wrote — precedence, the shape check, the helper, the tests are all still yours. Your probe table is still the only real-host verification this fix has; the suite skips Windows and mocks prove plumbing only. If you ever re-run it on that machine against `main`, I'd like to hear the result.

cemkizildag1907:
**Real-host verification against `main` (0.8.8, `0a09b96`), Windows + Claude Code desktop host.**

**1. The fallback path (the bit CI can't reach).** In a subprocess spawned by the host, `CLAUDE_CODE_OAUTH_TOKEN` is absent — the redaction we could never reproduce in the suite. With `REMEMBER_OAUTH_TOKEN` set to a real `setup-token` value and no host token present, `call_haiku` authenticated and the nested `claude -p` returned normally (`is_skip=False`, got the sentinel reply). So the config-based fallback recovers auth end-to-end on the host that withholds the token — same outcome as the original registry probe, now via consent-based config.

**2. #184, on the real host.** A malformed `REMEMBER_OAUTH_TOKEN` (9 chars) is now written to the daily log — `WARNING: ignoring REMEMBER_OAUTH_TOKEN — not a plausible OAuth token (… got 9 chars); the nested CLI will run unauthenticated …` — with source and length, never the value. It no longer vanishes; "rejected" is now distinguishable from "never set". (#190's "tokens … unknown" line showed up on the same failed call too.)

Net: does what it set out to on the real host. Thanks again for the review — the consent redesign is genuinely better, and #184 closed the silent-refusal gap our own docs claimed was already closed.

fdaviddpt:
Thanks for going back to the real host — that's the half the suite structurally can't reach, and it's the only part of this that produced new information. Everything CI can see was already green before the fallback existed; the redaction only happens in a subprocess the runner never spawns.

Two things worth recording from your run.

The fallback path has no automated guard and won't get one, so the only evidence it works was one manual run in a comment thread. I've opened #197 to turn your procedure into something the repo owns — including the check that `CLAUDE_CODE_OAUTH_TOKEN` really is absent from the child, since without it the test can pass for the wrong reason.

On #184: the silent refusal was the worse half of that bug, and it was ours. The docs claimed a configuration-time failure we'd never implemented, so the code and the promise disagreed and the promise won the argument in everyone's head. Good to have confirmation the warning carries source and length and never the value. #190 firing on the same failed call is right too — "unknown" beats an implied zero.

PR #181, opened 2026-07-26
fix(save): make the stale-lock takeover atomic

fdaviddpt:
Follow-up to #168, which closed one half of the lock-ownership problem and left the other open.

## The bug

The stale-lock takeover branch was a plain overwrite with no compare-and-swap:

```sh
log "lock" "stale lock (PID $LOCK_PID dead), taking over"
echo $$ > "$LOCK_FILE"
HAVE_LOCK=true
```

Two processes that both observe the same dead PID both take over, both set `HAVE_LOCK`, and both proceed — two concurrent saves, two Haiku calls, two position writes. That is the corruption #168's own description names, reached by the stale path rather than the contention path.

Reproduced with two real processes against a pre-seeded dead-PID lock in **4 of 5 runs**.

## The fix

Unlink, re-create under `noclobber` so only one contender can create the file (the loser skips this cycle), then re-read and stand down unless it holds our own PID. Whoever writes last wins, everyone else exits. Never two holders.

**Not a true CAS, and the code says so.** Between the verify and the first real work another takeover could still clobber the file — microseconds wide, confined to this branch. `mkdir` has real atomic-create semantics and is the honest long-term fix, but it changes the lock representation, so it is not this PR. Worth its own issue.

## Second defect, found while testing

`LOCK_PID=$(cat "$LOCK_FILE" 2>/dev/null)` is a plain top-level assignment, so under `set -e` its exit status is `cat`'s — and the lock file can be gone by the time it is read. Pre-existing, but **this PR's own `rm -f` makes the window easy to hit**: the loser exited 1 with an ERR-trap log in ~2 of 12 runs where it should have skipped cleanly with 0.

It never let a second process through (`HAVE_LOCK` is not yet set, so `cleanup()` is a no-op on the lock), but it turned a normal outcome into a logged failure. Guarded with `|| true`.

## Verification

- Test spawns two real `save-session.sh` processes against a dead-PID lock, 8 iterations per run, asserting at most one runs `extract`/`call-haiku`.
- **Revert-proved:** restore the three pre-fix lines and it fails at iteration 0 — `both processes proceeded past the stale lock takeover (extract ran 2 times)`, with both call logs showing the full pipeline.
- 5 consecutive runs green (40 race iterations) with strict exit-code assertions restored after the `|| true` guard.
- ~100+ race iterations during investigation, never a double winner.

## Checked and cleared

`[ "${REMEMBER_DEBUG:-1}" = "1" ] && log ...` immediately before `exit 0` looked like a `set -e` hazard. It is not — bash exempts a short-circuited final member of an `&&` list. Verified empirically for all three occurrences with `REMEMBER_DEBUG=0`.

Co-Authored-By: Max <noreply>

### DIFF
--- a/scripts/save-session.sh
+++ b/scripts/save-session.sh
@@ -90,13 +90,48 @@ trap cleanup EXIT
 
 # --- Lock (atomic via noclobber) ---
 if ! ( set -o noclobber; echo $$ > "$LOCK_FILE" ) 2>/dev/null; then
-    LOCK_PID=$(cat "$LOCK_FILE" 2>/dev/null)
+    # `|| true` is load-bearing under set -e. This is a plain top-level
+    # assignment, so its exit status is cat's, and the lock file can be gone by
+    # the time we read it — the holder finished and cleaned up, or another
+    # taker-over unlinked it in the takeover branch below. Without the guard
+    # the ERR trap fires and the script exits 1 ("FAILED at line ...") where it
+    # should simply skip this cycle with 0. An empty LOCK_PID then fails the
+    # kill -0 test and falls through to takeover, which is the right reading of
+    # a lock file that no longer exists.
+    LOCK_PID=$(cat "$LOCK_FILE" 2>/dev/null) || true
     if kill -0 "$LOCK_PID" 2>/dev/null; then
         [ "${REMEMBER_DEBUG:-1}" = "1" ] && log "lock" "locked by PID $LOCK_PID, skipping"
         exit 0
     fi
-    log "lock" "stale lock (PID $LOCK_PID dead), taking over"
-    echo $$ > "$LOCK_FILE"
+    # `echo $$ > "$LOCK_FILE"` is a plain overwrite with no compare-and-swap,
+    # so two processes that both observe the same dead PID both "take over"
+    # and both set HAVE_LOCK — the concurrency #168 closed on the acquisition
+    # path, still open on this one. Two saves, two Haiku calls, two position
+    # writes. Reproduced with two real processes against a pre-seeded dead-PID
+    # lock in 4 of 5 runs.
+    #
+    # Unlink, then re-create under noclobber so only one contender can make the
+    # file; the loser skips this cycle. That still leaves the window where a
+    # second taker-over unlinks the lock we just created, so re-read it and
+    # stand down unless it holds our PID. Whoever writes last wins and everyone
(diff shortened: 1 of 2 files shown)


fdaviddpt:
Correcting this PR: **the central claim in the description is false.** Adversarial review disproved it empirically.

I wrote "whoever writes last wins and everyone else exits — never two holders." That holds for exactly two contenders, which is all my test exercises. Beyond that it breaks:

| contenders | multi-winner rounds | max simultaneous winners |
|---|---|---|
| 2 | 0/40 | 1 |
| 4 | 1/40 | 2 |
| 5 | 2/40 | 2 |
| 8 | **11/40 (27.5%)** | 2 |
| 10 | 7/30 | **4** |

The verify step (`[ "$(cat "$LOCK_FILE")" != "$$" ] && exit 0`) is a one-time snapshot, not a held lock. Once a process passes it and sets `HAVE_LOCK=true`, it never re-checks — so a later contender's `rm -f` + recreate clobbers an already-verified winner's lock while it is still mid-flight.

Worse, and I missed it entirely: `cleanup()` gates on the `HAVE_LOCK` boolean alone, never on whether `LOCK_FILE` still contains its own PID. A displaced winner therefore deletes whatever is currently there on exit — including the lock of a process that legitimately acquired it afterwards. That is exactly the "non-owner deletes the lock" bug #168 fixed, reintroduced through the path I added.

And `>2` concurrent invocations is not a contrived load here: `PostToolUse` fires per tool call, and batching several tool calls per turn is normal in this product.

**Closing this rather than patching it.** The approach is wrong at the root: `rm -f` followed by create is not atomic and cannot be made atomic by adding verification after it. Any fix built on that sequence is a narrower window, not a lock.

The real fix uses primitives that actually are atomic:

- `mkdir` for acquisition — succeeds for exactly one process, fails for all others
- `mv` (rename) to break a stale lock — two processes both attempting to rename the same directory means exactly one succeeds and the loser gets ENOENT, so the takeover itself becomes single-winner

That changes the lock representation, so it belongs in its own change with its own tests at N=2, 4, 8 and 10 — the concurrency levels that exposed this.

The `|| true` guard on `LOCK_PID=$(cat ...)` is a genuine independent fix and I will carry it forward. Note for whoever picks that up: reverting it is caught by the existing test only ~25% of the runs, so it needs its own deterministic test rather than relying on race timing.

Thanks to the review for pushing past the two-process case. I would have shipped this.

PR #188, opened 2026-07-26
fix(ndc): record what a refused compression cost

### LINKED ISSUE #180: NDC reject path spends Haiku tokens that never reach log_tokens
`scripts/save-session.sh` NDC block: `log_tokens "ndc" "$TK_IN" "$TK_OUT" "$TK_CACHE" "$TK_COST"` fires only inside the success branch (`elif [ -n "$NDC_TEXT" ]`).

Since #178 the block has a third outcome — `IS_SKIP`/`IS_REJECTED` true, meaning the model returned a refusal or a clarification. That call still went out, still spent input and output tokens, and still cost money. None of it is recorded.

Pre-existing: `log_tokens` was gated the same way before #178 (`git show 3da8891:scripts/save-session.sh`). #178 only made the untracked path reachable more often, by giving rejections their own branch instead of letting them fall through into the append.

**Why it matters beyond tidiness:** a model that starts refusing compression produces a run where memory stops growing *and* reported cost drops to zero, which reads like "nothing happened" rather than "something failed repeatedly and was paid for". That is the same invisibility class #178 was about.

**Fix:** move the `log_tokens` call above the branch, or repeat it on the reject path. Worth checking the summarize path (Step 6) for the same shape while in there — a rejected summarize call is billed too.

Found during adversarial review of #178, out of scope for that PR.

fdaviddpt:
Closes #180.

`log_tokens` sat inside the success branch of the NDC block. Since #178 that block has a third outcome — the model answering with a refusal or a clarification — and those calls went out, spent input and output tokens, and cost money that was never recorded.

The failure mode is the one #178 was about: a model that starts refusing compression gives a run where memory stops growing **and** reported cost drops to zero. That reads as "nothing happened" rather than "this failed repeatedly and was paid for".

The call is now accounted for before its verdict is read, because spending it is what already happened.

The issue suggested checking the summarize path in Step 6 for the same shape — it already logs before it branches, so it needs nothing.

## Test

`tests/test_ndc_reject_gate.py` gains a case that drives a rejected NDC call with real token counts and asserts the numbers reaching the log are the ones the call spent, not merely that some line exists. The reject-gate stub gains `STUB_NDC_TOKENS` for that; unset, it emits zeros exactly as before, so no existing test changes behaviour.

Red against pre-fix code, green after. Full suite: **735 passed, 41 skipped**.


fdaviddpt:
Review point taken — the entry did read as though every NDC call is now accounted for, and it is not.

A call that times out or exits non-zero raises inside `call_haiku` before any token counts exist, so there is nothing to log on that path: `cmd_call_haiku` catches the `RuntimeError` and exits 1 without ever emitting `TK_*`. That is a real hole of the same shape, one branch over, and it needs either parsing usage out of the JSON failure payload or logging the attempt with an explicit unknown cost — neither of which belongs in this diff.

CHANGELOG narrowed to the outcomes this actually covers, and the remaining gap is tracked in #190 rather than implied away.

Issue #159, opened 2026-07-25
features.ndc_compression is documented but read nowhere — setting it false does nothing


fdaviddpt:
## Description

`features.ndc_compression` is documented in the README as a supported option, but the string appears nowhere in the code. Setting it to `false` has no effect: NDC compression runs anyway.

```
$ grep -rn "ndc_compression" scripts/ pipeline/ config.json README.md
README.md:223:| `features.ndc_compression` | `true` | Enable hourly compression of daily files |
```

One documented row, zero readers.

## Reproduction

1. Set `"features": {"ndc_compression": false}` in the config.
2. Run a save that produces an entry.
3. Watch the log.

```
[write] appended: ## 23:41 | main
[ndc] now.md → today-2026-07-25.md
[ndc] running (PID 86806)
```

Compression runs regardless. The only thing that actually holds it back is the `tmp/last-ndc.ts` cooldown marker together with `cooldowns.ndc_seconds` — `scripts/save-session.sh` Step 8 gates on the marker alone and never consults the feature flag.

## Why it matters

Someone who wants their `now.md` left alone — to read it raw, to debug what the summarizer is producing, or because they run their own compression — sets the documented option, sees it ignored, and has no signal as to why. The workaround (a far-future `ndc_seconds` plus a hand-written marker file) is not discoverable from the docs.

It also makes tests around save output awkward: anything asserting on `now.md` contents has to defeat the background compression by hand, since the documented brake does nothing.

## Suggested fix

Gate Step 8 on the flag alongside the cooldown, matching the existing `config ".key" default` pattern used elsewhere in the script:

```bash
if [ "$(config '.features.ndc_compression' true)" != "true" ]; then
    RUN_NDC=false
fi
```

Defaults to `true`, so nothing changes for anyone who has not set it.

Worth a test that sets the flag to `false` and asserts no `[ndc]` line is logged — the failure mode here is silence, which is how it went unnoticed.

## Notes

Found while writing tests for #136: a test asserting a rejected entry never reached memory passed against *unfixed* code, because compression had already moved the junk out of `now.md` and into `today-*.md`. The flag was set to `false` in that fixture and did nothing.

PR #189, opened 2026-07-26
fix(config): wire the option that did nothing, document the ones that did

### LINKED ISSUE #176: Config contract gaps: a dead documented option, an undocumented one, mismatched defaults
Audit of the README config table against what the code reads.

**1. `debug` is dead.** Documented (default `false`, "Verbose logging for cooldowns and locks") and shipped in `config.example.json`, but `.debug` is passed to `config()` **nowhere**. The actual verbose logging is gated on the `REMEMBER_DEBUG` env var. Setting `"debug": true` does nothing. Same class as #159.

**2. `git_backup.allow_remote_change` is undocumented** — in neither README nor `config.example.json`, though the code reads it. It is security-relevant: `true` bypasses the changed-remote abort, letting memory push to a remote that changed under you. Discoverable only by reading source.

**3. `REMEMBER_DEBUG` means different things in different scripts.** `save-session.sh` uses `${REMEMBER_DEBUG:-1}` (verbose by default); `50-git-backup.sh` uses `${REMEMBER_DEBUG:-0}` (quiet). The README documents a single default of `1`, which git-backup contradicts.

**4. `thresholds.consolidate_max_bytes`** is read by `run-consolidation.sh` and explained in `config.example.json`, but missing from the README table (its sibling `extract_max_bytes` is there).

Worth a test that walks the README table and asserts each key is read somewhere — the #159 class is "documented option that does nothing", and it has now happened twice.

fdaviddpt:
Closes #176.

## `debug` did nothing

It was in the README table, shipped in `config.example.json`, and passed to `config()` **nowhere**. The real switch was the `REMEMBER_DEBUG` env var — which someone configuring the plugin through `config.json` has no obvious way to set. Same class as #159, where documented booleans could not be switched off.

Wired now, through a shared `debug_enabled` helper. Precedence: env var, then the config option, then **the caller's own default**.

That last part is the interesting one. `save-session.sh` was verbose unless told otherwise; `50-git-backup.sh` was quiet unless told otherwise; the README documented only the first, so its "default `1`" had always been contradicted by the git-backup hook. Collapsing them into one shared default would have silently changed log volume for every existing install in one direction or the other — so each keeps its own, and setting the option overrides both. Which is what setting it was supposed to do.

## Documentation that had drifted the other way

- `thresholds.consolidate_max_bytes` — read by `run-consolidation.sh`, in the example config, absent from the README table. Its default is **600000**; a first draft of this PR wrote 200000 from the issue text and was wrong until I read the code.
- `model` and `reject_pattern` — read, shipped, and documented **only** as env vars, though `config.json` is the source of truth for both.
- `git_backup.remote` / `branch` / `gpg_sign` — documented, read, missing from `config.example.json`.

`model` and `reject_pattern` are the new test's finding, not the issue's.

## The test

`tests/test_config_contract.py` compares three sides — the README config table, `config.example.json`, and the keys the code actually passes to `config()` — in **both** directions, and drives the `debug` switch through bash rather than asserting the key is merely mentioned somewhere.

One thing it taught me while being written: `timezone` is documented and read but deliberately **not** shipped in the example config, because a `"timezone": ""` people copy is the UTC landmine an earlier fix removed — and `test_log_sh.py` already asserts it stays out. My first version added it back and broke that test. It is now named as a deliberate exception with the reason attached, which is the point of the exercise: encode the intent, don't relitigate it.

Eight of its ten tests go red against pre-fix code. Full suite: **744 passed, 41 skipped**.


fdaviddpt:
Good catch, and it is the sharper version of the point the PR itself makes.

`"debug": false` had sat in `config.example.json` doing nothing for as long as the option was unread. Wiring the option up turned that dead line live — so anyone copying the example file would have had `save-session.sh` go quiet, which is exactly the silent default change the commit message argues against. Same landmine class the suite already names for `timezone`, and I walked straight into it while writing the rule that describes it.

Removed the key, added `debug` to `_DELIBERATELY_NOT_SHIPPED` with the reason attached, and added a test asserting it stays out — so the next person to "complete" the example file has to read why it is incomplete on purpose.

On the two non-blocking notes: the ~20ms is real but lands in the backgrounded save path rather than the tool-call critical path, as you measured, so I have left it. The fail-through-to-caller-default for garbage values is deliberate — `features.*` fails closed because a broken feature flag should disable the feature, while a broken verbosity flag should not decide log volume on the user's behalf — but you are right that it reads inconsistently, so the reason is now in the comment.

Full suite 745 passed, 41 skipped.

Issue #139, opened 2026-07-24
save-session: header time is validated by format only — the summarizer can stamp in-transcript times, scrambling entry order


bonyohana:
## Observed (production, 0.8.3 via marketplace; mechanism re-verified on main @ 7db7866)

Daily memory files showed out-of-order headers: `## 18:30` written at 05:10, `## 18:15` written at 06:08, `## 18:33` written at 08:34 (wall-clock times from the plugin's own log; headers from the resulting `today-*.md`). Entries sort/read as if written hours apart from their true order.

## Mechanism

`prompts/save-session.prompt.txt` injects a concrete `{{TIME}}` and (since the recent hardening) tells the model to copy it verbatim — but the model sees the whole transcript, which (with the UserPromptSubmit hook active) contains other timestamps, and sometimes stamps one of *those* instead. The Step-5b validator (`scripts/save-session.sh:192`) only regex-checks the header **format** (`^## HH:MM |`), never the **value**, so a well-formed wrong time passes silently:

```bash
if ! echo "$FIRST_LINE" | grep -qE '^## ([0-9]{2}:[0-9]{2}|[0-9]{1,2}:[0-9]{2} (AM|PM)) \|'; then
```

(#136 covers malformed output being written anyway; this is the complementary case — well-formed but wrong-valued, which no warning ever fires for.)

## Suggested fix

The script already computed the time it put into `{{TIME}}`. Compare the parsed header time against it (± a small tolerance for the call duration) and, on mismatch, **rewrite the header to the known-correct time** rather than warn — this is the one field the pipeline knows better than the model. One-line normalization after parse-haiku; kills the whole fault class.

PR #191, opened 2026-07-26
fix(rules): one entry-header pattern, one slug, one threshold

### LINKED ISSUE #177: Remaining duplicated rules free to drift (entry-header regex, run-tests slug, test _slug copies)
Remaining copies of rules that exist in more than one place. None disagree *today*; all are free to drift, which is how the slug (5 copies), the position readers (5) and the projects path (5) each drifted before.

**1. Entry-header regex.** `pipeline/consolidate.py:_ENTRY_HEADER` accepts 24h only, no trailing pipe. `save-session.sh` accepts 12h AM/PM and requires `" |"`. Normally reconciled because the header is rewritten to 24h — **except** the re-validation fallback added in #139, which keeps the model's original line when splitting fails. If that fires on an AM/PM header, `consolidate.py` will not recognise the entry at all.

**2. `scripts/run-tests.sh`** builds the session dir inline rather than sourcing `detect-tools.sh`. Safe only because `mktemp -d` yields ASCII.

**3. Three test files** (`test_post_tool_cooldown.py`, `test_external_data_dir.py`, `test_worktree_memory.py`) each carry their own `_slug` reproduction of the naive rule. They agree only because no fixture uses a non-ASCII path.

**4. `delta_lines_trigger` default (50)** lives in `post-tool-hook.sh`; `test_post_tool_cooldown.py` hardcodes 60 lines and relies on the shipped default. *Raising* the default fails the test loudly; *lowering* it goes unnoticed.

Suggestion: one shared slug implementation reachable from bash, Python and tests (see also #158 and the astral mismatch), and a single source for thresholds that tests import rather than restate.

fdaviddpt:
Closes #177.

Four rules that existed in more than one place. None of the copies disagreed when the issue was filed — which is the argument for closing them now rather than after the next silent outage, the way the slug (five copies), the position readers (five) and the projects path (five) each went.

## The one that could already bite

`save-session.sh` accepted `## 9:30 AM | main`. `pipeline/consolidate.py` accepted 24-hour only.

They agreed in practice because the header is rewritten to 24h before it is written — **except** through the re-validation fallback added in #139, which keeps the model's original line when a rewrite would malform it. An entry reaching memory that way reads as prose to consolidation, and a consolidation carrying only such entries is discarded as "purely conversational".

The pattern now lives in `pipeline/entry_header.py`, and the shell asks for it:

```sh
ENTRY_HEADER_ERE=$($PYTHON -m pipeline.entry_header --ere entry)
```

with the literal kept as a fallback — a failed lookup must not leave `grep -E` matching an empty pattern, which matches everything.

## The other three

- **`run-tests.sh`** built the session directory with an inline `sed`, safe only because `mktemp -d` yields ASCII. It now calls `session_dir_slug`.
- **Three test files** each reproduced the naive slug. They agreed only because no fixture used a non-ASCII or over-200-character path — the same "true today" that let the shipped copies drift. They import `pipeline.slug` now.
- **`delta_lines_trigger`** — the cooldown tests inherited the shipped default instead of stating it. *Raising* it failed them loudly; *lowering* it changed nothing and told nobody. The fixture writes the threshold it depends on, and a test asserts the hook's built-in default and `config.example.json` still agree.

## Tests

`tests/test_shared_rules.py` fails when a second definition **appears**, not after it has drifted. It also checks the header pattern is portable between Python's `re` and `grep -E` — that portability is the price of having one definition two engines read, so it is worth asserting rather than assuming.

It found a fifth suspect on its own, which turned out to be a comment and a docstring quoting the old rule rather than a copy of it; the detector now ignores prose.

Three of its tests go red against pre-fix code, including the 12-hour entry driven through `consolidate.py`'s own matcher rather than through the new module — my first version tested the new code and would have passed against the bug.

Full suite: **811 passed, 41 skipped**.

## Note

The lock half of #177 (`save.lock` and `consolidation.lock` as two copies of one rule) was closed by #185.


fdaviddpt:
All four addressed, and two of them were things I had argued against doing while doing them.

**The fallback was an unchecked copy.** Exactly right — deliberate as a fallback, but nothing tied it to the source of truth, so `TIME_ERE` could change and it would silently go stale. A test now asserts the literal in `save-session.sh` equals `ENTRY_HEADER_ERE`.

**The coverage loss is real and I caused it.** Writing an explicit threshold in the fixture fixed the silent-inheritance problem and removed the only thing exercising the hook's own default. Two tests now run it with no config at all, 60 lines and 40, either side of 50.

**The multi-line asymmetry** — `is_entry_header` was compiled without `re.MULTILINE` while `grep` anchors per line, which makes "one pattern, two engines" untrue for any caller passing more than one line. Both compile `MULTILINE` now and the parity test covers multi-line input.

**The widened accept surface** I have closed rather than documented: `ANY_HEADER_ERE` now requires the identity pipe after a time, so `## 9:00 AM standup notes` is not a header. Every real entry has the pipe because `save-session.sh` refuses to write one without it, and a negative test drives your exact refusal example through `_is_valid_consolidation`.

On the nitpicks: you are right that it is four tests red, not three — corrected. And the slug detector does only catch literal copy-paste; that was the historical failure mode, but the docstring overstated it, so it now says what it checks.

Full suite 818 passed, 41 skipped.

Issue #157, opened 2026-07-25
session_dir_slug does not truncate at 200 chars + hash, so deep project paths silently never save


fdaviddpt:
## Description

`session_dir_slug()` never truncates. Claude Code does: past 200 characters it keeps the first 200 and appends a hash of the path. So for any project whose slugged path exceeds 200 characters, the computed slug does not match the directory Claude Code created, `SESSION_DIR` points at nothing, the hook finds no transcript, and memory silently never saves.

This is the same failure mode as #144, reached by a different route — long path instead of non-ASCII path — and it is **not** fixed by #156.

## Reproduction

Confirmed on a real session, not inferred. A project path 262 characters long:

```
$ cd /very/deep/…/gggggggggggggggggggg     # 262 chars
$ claude -p 'reply with the single word ok'
ok

$ ls -t ~/.claude/projects/ | head -1 | awk '{print length, $0}'
207 -private-tmp-…-dddddddddddddddddddd--on930v
```

Claude Code wrote a **207**-character directory. `session_dir_slug` returns the full **262**-character slug. No overlap, so nothing is ever found.

## The algorithm

From the shipped CLI bundle:

```js
var jXA = 200;

function JXA(A) {
  let q = A.replace(/[^a-zA-Z0-9]/g, "-");
  if (q.length <= jXA) return q;
  let K = 0;
  for (let _ = 0; _ < A.length; _++) K = (K << 5) - K + A.charCodeAt(_), K |= 0;
  return `${q.slice(0, jXA)}-${Math.abs(K).toString(36)}`;
}
```

So: djb2-ish rolling hash over the **original** path (not the slug), `charCodeAt` so UTF-16 code units, forced to signed 32-bit by `|0` each round, then `Math.abs(...).toString(36)`.

I reimplemented that and it reproduces the observed directory name exactly (`…dddddddddddddddddddd--on930v`), so the algorithm above is verified rather than read-and-hoped.

## Suggested fix

Extend `session_dir_slug` in `scripts/detect-tools.sh`: if the slug exceeds 200 characters, emit `${slug:0:200}-${hash}`. The hash needs 32-bit signed wraparound and base36, which is awkward in pure bash — `$PYTHON` is already a hard dependency of every hook, and this branch is only reached for long paths, so paying a subprocess there costs nothing on the common path.

Worth pinning with a test that builds a >200-character path, since this class of bug is invisible until someone's checkout happens to be deep.

## Notes

[...]

Issue #144, opened 2026-07-25
session_dir_slug() mis-slugs non-ASCII (CJK) paths under byte-wise sed, silently disabling saves


shenk-ana:
## Description

`.remember/` never produces `now.md` / `today-*.md` / etc. for a project whose path contains non-ASCII (e.g. CJK) directory components — the save pipeline silently never runs, with no error anywhere.

## Reproduction

Steps to reproduce:

1. Use a project whose path contains a non-ASCII directory component (e.g. `D:/foo/プロジェクト/bar`).
2. Start a Claude Code session in that project and do enough tool calls to exceed the save threshold (default 50 JSONL lines).
3. Observe `.remember/logs/*.log` and `.remember/` itself.

## Expected behavior

Once the delta threshold is exceeded, `save-session.sh` should run in the background and eventually produce `now.md` (and later `today-*.md` via NDC compression).

## Actual behavior

`.remember/logs/*.log` only ever shows `[hook] post-tool: ...` trace lines — never `[extract]`/`[haiku]`/`[write]`. `.remember/` never gets a `now.md`. `hook-errors.log` stays empty. No indication anywhere that anything is wrong.

## Investigation

`post-tool-hook.sh` locates the current session's JSONL transcript via `SESSION_DIR="$HOME/.claude/projects/$(session_dir_slug "$PROJECT_DIR")"`, then does `[ -n "$LATEST_JSONL" ] || exit 0`. For a project path with non-ASCII components, `session_dir_slug()` (in `scripts/detect-tools.sh`) computes the wrong slug, so `SESSION_DIR` doesn't exist, `LATEST_JSONL` is never found, and the hook exits before ever reaching `save-session.sh` — with nothing logged to indicate why.

Root cause: `session_dir_slug()` ends with `sed 's/[^a-zA-Z0-9]/-/g'`, and on this host `sed` matches byte-by-byte rather than character-by-character unless `LC_ALL` is explicitly forced to a UTF-8 locale — even though `LC_CTYPE=C.UTF-8` is already set:

```bash
$ locale
LANG=
LC_CTYPE="C.UTF-8"
...

$ echo "D:\foo\プロジェクト\bar" | sed 's/[^a-zA-Z0-9]/-/g'
D--foo--------------------bar   # 18 dashes — one per UTF-8 *byte* of the 6 CJK chars (6*3=18)

[...]

PR #193, opened 2026-07-26
fix(paths): normalise CLAUDE_CONFIG_DIR, and cover it where it can break

### LINKED ISSUE #169: CLAUDE_CONFIG_DIR is not path-normalised on Windows, unlike PROJECT_DIR

fdaviddpt:
Closes #169.

`claude_projects_dir()` concatenated `/projects` onto whatever the environment held. `PROJECT_DIR` is deliberately converted from Git Bash's POSIX form to the native Windows one before it is slugged; `CLAUDE_CONFIG_DIR` got nothing, so a natively-expressed value produced `C:\Users\x\.claude-alt/projects/<slug>`.

Whether that ever actually failed is still unknown — MSYS translates both forms in syscalls. The issue was filed unverified for want of a Windows machine, and it stayed open because the failure mode is the one this project keeps hitting: the session directory is not found, `ls` returns nothing, the pipeline no-ops. "Probably fine" is what #144, #166 and #157 all were before someone hit them.

Converted with `cygpath -u` when it exists, and stripped of trailing separators either way, so `C:\Users\x\.claude-alt\` cannot become `…//projects`.

## The half that matters more

The issue's real complaint is the one I could actually close: the only test covering this path is `skipif(win32)`, in line with the repo's convention for bash-subprocess tests. So it had **zero coverage on the only platform where it can fail**, and this class could only ever be caught by a user.

Two answers:

- **The bash branch runs on POSIX CI** against a stub `cygpath` that mimics the MSYS conversion. The branch is Windows-only; the logic is not, and pinning it does not require the platform.
- **The Python half runs unskipped on the Windows runner** — `_session_dir` needs no bash — across every form a Windows environment can hold: native, forward-slash, and both with trailing separators.

Four of the new tests go red against the unfixed function, including two POSIX ones: trailing slashes were not stripped before either, which is a small real bug on every platform rather than a hypothetical Windows one.

Full suite: **795 passed, 41 skipped**.


fdaviddpt:
All three taken, and the third one is a fair hit on the description rather than the code.

**The empty-root collapse is fixed.** `/` and `///` stripped to nothing and put `/projects` at the filesystem root; a lone `\`, which is a *relative* path, became an absolute one pointing somewhere else. Both keep their meaning now, with the degenerate inputs pinned by tests. You are right that silently reinterpreting a path is the exact shape this file's history is made of.

**The Windows-runner claim was wrong** and I have corrected it rather than softened it. That test exercises `pipeline/extract.py`, which this PR does not touch, and it passes against the unfixed code — so it demonstrates nothing about the change under review. The bash fix still has no Windows coverage; the description and the CHANGELOG now say that plainly.

**On the drift** — I have not unified the two builders, because I do not think they should be. The slug was unifiable since it is a *name*: one string, one right answer. This is a *path*, and the right form depends on which runtime dereferences it — bash under Git Bash wants POSIX, a native Windows Python wants native. Forcing one form would break the other consumer. The invariant that matters is "same directory", not "same string", and nobody has established it holds on Windows. A parity test covers the POSIX case where the forms coincide, and #194 tracks the Windows case with what would actually settle it.

Nitpicks noted: the newline handling asymmetry with `session_dir_slug` is real, and the UNC case in the stub does work by accident of ordering rather than intent — both worth a follow-up rather than widening this diff further.

Full suite 833 passed, 41 skipped.

PR #195, opened 2026-07-27
fix: close the three follow-ups from the review pass

### LINKED ISSUE #186: lib-slug.sh disagrees with the decoder on malformed UTF-8 mid-string, and Linux allows such paths
Found during adversarial review of #157/#158/#174 (PR pending). Narrow, but it is the exact class of bug that work exists to remove, so it should be written down rather than left in a comment.

## The divergence

`scripts/lib-slug.sh`'s UTF-8 byte table emits one dash per byte for a lead byte that is not followed by the continuations it requires. The real decoder folds that into a single replacement character by the maximal-subpart rule, which is one dash.

Reproduced:

```
$ printf 'ab\xe0\xa0cd' | ... session_dir_slug   ->  ab--cd
$ node: Buffer.from([...]).toString('utf8')      ->  ab�cd  ->  ab-cd
```

Same for a truncated 4-byte lead (`\xf0\x90` followed by ASCII).

## Why the existing comment understates it

`lib-slug.sh` says the only divergent shape is a truncated sequence at the **end** of the string, and that it "cannot arise from a path a filesystem would hand us: macOS enforces well-formed UTF-8, Windows paths come from UTF-16."

Both halves are too generous:

- It is not only end-of-string. It happens wherever a partial multi-byte lead is followed by a non-continuation byte, anywhere in the path.
- Linux is not covered by that argument. ext4 and friends do not enforce UTF-8 in filenames — every byte except `/` and NUL is legal. A legacy-encoded (Latin-1, Shift-JIS) or corrupted directory name gets there without anything exotic.

## Consequence

On Linux, a small class of non-UTF-8 project paths still resolves to a slug Claude Code never created: no transcript found, nothing saved. Since #156 the hook logs a warning when the session directory is missing, so it is not fully silent — but it is the same shape as #144/#166.

## Fix

Expressing maximal-subpart in the sed table is not possible — it needs the decoder's state machine. Options, roughly in order of appetite:

- Have the shell side call `pipeline/slug.py` for any path containing a byte >= 0x80 that fails a cheap well-formedness check, the way it already calls out for the over-200 hash. Costs a subprocess only for paths that are already unusual.
- Or accept the divergence and narrow the comment to the truth, which is what the pending PR does.

Either way the comment should not claim a guarantee the table does not provide.

### LINKED ISSUE #190: Tokens spent on a Haiku call that fails or times out are never accounted for
Found during review of #188 (which fixes the *refusal* case of #180). Separate path, same shape.

## What is not recorded

When the `claude` subprocess itself fails — a timeout against the 180s NDC budget, a mid-stream API error, a non-zero exit — `pipeline/haiku.py::call_haiku` raises `RuntimeError` **before** returning a `HaikuResult`. `cmd_call_haiku` (`pipeline/shell.py`) catches it, prints to stderr and exits 1, so no `TK_*` variables are ever emitted and the caller's error branch logs the message with no cost attached.

Tokens can genuinely have been spent by then: a client-side timeout aborts a call the API has already been billing, and a mid-stream failure has already consumed input.

## Why it matters

It is the invisibility class #178 and #180 are both about. A run where the model times out repeatedly shows errors in the log and **zero reported cost**, which reads as "it failed for free". #188 closed that for refusals; this is the same hole one branch over.

## Why #188 did not fix it

There is nothing to log. The counts do not exist on that path — `call_haiku` never gets far enough to parse a usage block. Closing this needs one of:

- **Parse usage out of the failure payload.** `--output-format json` reports errors as JSON on stdout (that is what #129 was about), and that payload may carry a usage block. If it does, `_failure_detail` already reads it and could surface counts alongside the message.
- **Log the attempt without counts** — a line saying a call was made and its cost is unknown. Weaker, but "unknown" is honest and zero is not.

The first is preferable if the payload actually carries usage; worth checking against a real timeout before designing for it.

## Scope note

#188's CHANGELOG entry has been narrowed to say it covers the cases where a verdict came back, rather than implying every NDC call is now accounted for.

### LINKED ISSUE #194: The projects path is built twice, and on Windows the two forms differ
Found during review of #193.

Two implementations build `<config>/projects`:

- `scripts/lib-slug.sh::claude_projects_dir` — for the hooks, which are bash
- `pipeline/extract.py::_session_dir` — for the extractor, which is Python

With `CLAUDE_CONFIG_DIR='C:\Users\dev\.claude-alt'` they produce:

```
bash:   /c/Users/dev/.claude-alt/projects
python: C:\Users\dev\.claude-alt/projects/-p
```

Different strings for the same input.

## Is it a bug?

Not obviously. Each consumer needs the path in the form its own runtime can open: bash under Git Bash wants the POSIX form, and a native Windows Python wants the native one. Both may well name the same directory, which is all that is required.

But nobody has established that. It depends on which Python runs — native Windows Python tolerates mixed separators through the Win32 API; an MSYS Python would not — and #169's whole lesson is that "MSYS probably handles it" is where this class of bug hides.

## Why it is not simply unified

The slug was unified (#157/#158/#174) because it is a *name* — one string, one right answer. This is a *path*, and the right answer depends on the runtime dereferencing it. Forcing both to one form would break one of the two consumers.

The invariant to enforce is therefore not "identical strings" but "same directory".

## What would settle it

- A Windows-runner test that writes a transcript into the directory the Python side computes and asserts the bash side finds it, with `CLAUDE_CONFIG_DIR` set natively. That is the only thing that answers the question rather than reasoning about it.
- Failing that, one documented statement of the invariant next to both implementations, so the next person to touch either knows what must hold.

#193 added a parity test for the POSIX case, where no `cygpath` exists and the two forms coincide. The Windows case is what remains.

fdaviddpt:
Closes #190. Closes #186. Closes #194.

All three came out of adversarial reviews of the previous batch. Each was filed rather than fixed in place because it was a different question from the PR it surfaced in; this closes all three.

## #190 — a call that fails still spent tokens

Every accounting path in the pipeline hangs off a returned result, and a timeout or non-zero exit returns none. So a run where the model timed out repeatedly showed errors in the log and **zero** reported cost — "it failed for free". It did not: a client-side timeout aborts a call the API has already been billing, and a mid-stream error has already consumed input.

Usage now comes out of the JSON the CLI writes to stdout on failure when it carries any, and is reported as explicitly *unknown* when it does not. Zero is a lie; unknown is not.

## #186 — malformed UTF-8 disagreed with the decoder

The sed byte table is exact for well-formed UTF-8 and gives one dash per byte for anything else, where the real decoder folds an ill-formed sequence into a single replacement character. macOS enforces well-formed UTF-8 and Windows paths come from UTF-16 — but Linux enforces nothing, so a legacy-encoded or corrupted directory name resolved to a slug Claude Code never created.

Those paths now go to `pipeline/slug.py`, which owns a real decoder. The cost is bounded deliberately: a path pays a subprocess only if it is **both** non-ASCII **and** actually ill-formed, and an ASCII path never reaches even the `iconv` check — this function runs on every tool call, and trading that for a rare correctness fix would be a bad deal.

One thing that took a try to notice: `argv` arrives surrogate-escaped, so the first version of this delegated to Python and Python counted the bad bytes *exactly the way sed did*. The CLI now decodes back through the filesystem encoding first, which is what actually applies the maximal-subpart rule. Without that line the whole delegation buys nothing — and it looked like it worked.

## #194 — do the two builders name the same directory?

`lib-slug.sh` hands bash a POSIX path; `extract.py` hands Python a native one. On Windows those are different strings, and I argued in the last review that identical strings is the *wrong* invariant — each consumer needs the form its own runtime can open. The right invariant is that both name the same directory, and nobody had established that.

So the test creates the directory from the Python side and asks bash, through its own builder, whether it is there. On the Windows runner that is an answer rather than an argument, which is the whole complaint #169 was filed with.

## Tests

Three go red against pre-fix code for #190, four for #186 — including one asserting that well-formed paths never shell out, because a performance regression traded for a correctness fix is still a regression.

Full suite: **861 passed, 41 skipped**.


fdaviddpt:
Your blocker was right, and it was already fixed by the time the verdict landed — I benched my own claim while you were reviewing and found the same thing. Which is awkward for you: you reviewed `52be04f`, and the checkout you were reading was being edited underneath you. That is my fault and I have taken the lesson (review a committed SHA in a copy, never a worktree I am still working in).

Where it ended up:

- The check is now gated on `$OSTYPE` — a bash variable, not a fork — so it runs on **Linux only**, which is the one platform that can produce an ill-formed filename. macOS and Windows pay nothing at all. `REMEMBER_UTF8_STRICT=1` forces it on, because a fix nobody can exercise off Linux is a fix nobody maintains.
- Your nitpick about DEL and the C0 controls is fixed too — the glob now tests for a byte ≥ 0x80 rather than "outside space..~".
- And that change caught a much worse bug than the one it was for: `$'\\000'` expands to **nothing**, because bash cannot hold a NUL in a string. So `*[!$'\\000'-$'\\177']*` degenerated into `[!-\\177]` — "any character except `-` and DEL" — matching essentially every path. It reads as the tightest possible test and was the loosest one. Every ASCII path forked `iconv`.

That last one was found by the fork probe, not by reading the glob, which is the whole argument for the probe: I replaced my Python-only marker with one that shadows **both** `iconv` and `python3` and delegates to the real binaries — a stub that merely fails changes the behaviour it is measuring, which is how the first version 'found' a well-formed path reaching the decoder.

Measured, 200 calls each:

| | ASCII | valid non-ASCII | ill-formed |
|---|---|---|---|
| macOS / Windows | 1.9ms | 1.9ms | 2.2ms |
| Linux | 1.9ms | 7.5ms | 74ms |

The CHANGELOG now states this rather than the version you correctly called false.

Full suite 863 passed, 41 skipped.

fdaviddpt:
You caught the important thing: I wrote a test to answer the previous review's "unfalsifiable claim" complaint, and the test was itself unfalsifiable. It passes against the pre-fix code. That is the same defect one level up, and your file-swap proof is exactly right.

But the fix is not decoration, and here is the evidence I should have put in the commit instead of a diagnosis:

| run | commit | macOS |
|---|---|---|
| 30235721355 | `1a5b032` (Linux gate) | fail |
| 30236170724 | `1780266` (`\001` glob) | fail |
| 30237326112 | `fa888af` (bench) | fail |
| 30237861167 | `6ec61de` (`LC_ALL=C`) | **pass** |

Same test, three consecutive red runs on `macos-latest`, green on the run that forced byte semantics and nothing else. So the change is empirically necessary.

What I cannot demonstrate is **why**. Collation was my explanation and it is the obvious candidate — this file already documents that trap for `[a-z]` in its sed program — but your 55-locale sweep under bash 3.2 says it does not reproduce there, and I have no bash ≥4 on this machine to test the other half of the hypothesis. The runner differs in some way I have not identified.

So I have corrected the claim rather than the code:

- the code comment now reads *"empirically necessary, mechanism unconfirmed"*, with the four run IDs so the next person can check rather than trust me
- the test docstring says it is defense-in-depth, that the four locales are **not** known to be the ones that differ, and "do not read a pass here as proof the underlying bug is gone"
- the CHANGELOG says the same

`LC_ALL=C` stays. It is free, it is correct regardless of mechanism, and the CI record says it matters.

Your `PYTHON=` nitpick is fixed too — the benchmark exits 2 instead of printing an empty table and exiting 0, which read as "no measurable cost".

Full suite 867 passed, 41 skipped.

Issue #140, opened 2026-07-24
last-save.json is single-slot — interleaved sessions clobber each other's saved positions, producing duplicate memory entries


bonyohana:
## Observed (production, 0.8.3 via marketplace; mechanism re-verified on main @ 7db7866)

The same span of one session ("99 exchanges (11 human)" in the extract log) was summarized twice, ~2.5h apart (06:06 and 08:33), yielding a near-identical triplet of entries in the daily file.

## Mechanism

`pipeline/shell.py:cmd_save_position` writes a **single-slot** store:

```python
json.dump({"session": session_id, "line": position}, f)
```

and `pipeline/extract.py:get_last_save_line` returns 0 whenever the stored session differs:

> Returns 0 if the file is missing, corrupt, **or belongs to a different session.**

So with two live sessions A and B alternating saves (multiple terminals, or a background/worktree session sharing the store): A saves (slot=A), B saves (slot=B), A's next save finds slot=B → resumes from 0 → re-extracts and re-summarizes A's entire span → duplicates. The dedup context (last memory entry) doesn't catch it when other entries landed in between.

## Suggested fix

Key positions by session: either `last-save-<session>.json` files, or one JSON object `{"<session_id>": line, ...}` with a small LRU cap (sessions are UUIDs; a dozen entries suffice). Read path change is one lookup; write path merges instead of replacing.

PR #201, opened 2026-07-27
feat: name the silent no-op, and a doctor to ask about it (#200)

### LINKED ISSUE #200: Silent no-op when the plugin is enabled mid-session: PostToolUse never registers, no error surfaced
Thanks for the plugin — the pipeline itself works well. Reporting a failure mode that is **silent**, which made it hard to notice: memory simply never gets written, with no error anywhere.

## What happened

On Windows (Git Bash), the plugin was installed and enabled mid-session in a long-running Claude Code session. Result: `~/.remember/<slug>/logs/` contained a single `session-start` line and nothing else. No memory files were produced for that project over many hours and hundreds of tool calls. `logs/hook-errors.log` was empty, so nothing indicated a problem.

## Root cause

`PostToolUse` (which drives `save-session.sh`) was **never registered**, because Claude Code loads hook definitions at session start and the plugin was enabled after that session began. The plugin was working exactly as designed — it simply was not wired in yet.

Invoking the hook manually confirmed the pipeline is healthy end to end:

```
post-tool: PROJECT_DIR=... PYTHON=python3 REMEMBER_DIR=...
save-session: ...
[extract] 2490 exchanges (154 human)
[haiku] calling (branch: main)
[write] appended: ## 18:00 | main
[ndc] now.md -> today-2026-07-26.md
[ndc] 312->267b (-14%)
```

So the fix is just "restart Claude Code" — but there is no way for a user to discover that.

## Suggestions

1. **Make the gap detectable.** The `SessionStart` hook already runs; it could check whether a `PostToolUse` registration is visible (or simply whether any save has ever occurred for this project) and print a one-line notice such as *"remember: PostToolUse not active in this session — restart Claude Code to enable capture."* One line at session start would have saved a full day of silent no-ops.
2. **Say it in the README install steps.** "Restart Claude Code after enabling, otherwise capture will not run until the next session" — currently a user reasonably assumes enabling is sufficient.
3. **Optional: a `doctor`-style command** reporting resolved paths, detected tools, whether hooks are active, and the timestamp of the last successful save. That single output would make this class of problem self-diagnosing.

Happy to test a fix on Windows/Git Bash if useful. Environment: Windows 11, Git Bash (MSYS), Python 3.14, Node 25, `remember` 0.8.3 via `claude-plugins-official`, external storage mode (`data_dir: ~/.remember/{slug}`) with git backup enabled.

fdaviddpt:
Closes #200. Reported by @fmanimashaun.

## The failure

Claude Code reads hook registrations at session start, so a plugin enabled part-way through a session has none of them wired for the rest of it. `PostToolUse` never fires, nothing is captured, and `hook-errors.log` stays empty because nothing *failed*. The reporter lost a day to it and found one `session-start` line as the only trace.

The plugin was working exactly as designed. It simply was not wired in — and there was no way for a user to discover that.

## What can't be done, and why

Suggestion 1 in the issue was to have `SessionStart` check whether `PostToolUse` is registered. That is not possible, and it's worth recording so nobody re-attempts it:

- **No introspection exists.** No env var, no file, no CLI subcommand exposes which hooks are active. `/hooks` is a read-only UI a hook script cannot invoke.
- **`SessionStart` cannot tell why it fired.** Its `source` field is only `startup|resume|clear|compact|fork`. A plugin-enable is indistinguishable from a fresh start.

So the gap is undetectable *while it is happening*. What remains is detecting it afterwards.

## What is done

**Detection.** The signature is exact: a session where `SessionStart` ran and `PostToolUse` never did. `PostToolUse` records the session id it saw; the next `SessionStart` compares that against the previous session's id and, if that session used tools at all, leaves a notice.

The zero-tool guard is load-bearing: a conversation with no tool calls produces no `PostToolUse` either, and warning about it would be crying wolf.

Identity rather than timestamps — the first cut compared mtimes and a test caught it calling a **healthy** session broken, because bash 3.2's `-nt` resolves to the second and the sign of life landed inside the same one as the stamp.

**Delivery.** Through `systemMessage`, the only hook output the *human* sees. `additionalContext` reaches the model alone, which is the shape of the original bug and would have been the wrong channel. The ordinary path still prints exactly what it always did; only the notice case switches to JSON, and a test pins that so context injection cannot change silently.

**`/remember:doctor`.** Resolved paths, detected tools, storage mode, whether the session dir Claude Code created matches the computed slug (#144), the last successful save, and whether `PostToolUse` has ever fired. `OK`/`WARN`/`FAIL` per line, one `VERDICT`.

**README.** A restart note in the install steps, and a section on the diagnostics command.

## Two bugs its own tests found

1. An install predating the new `capture-alive` marker was reported as `FAIL PostToolUse has never fired` — while the next line printed a successful save. Self-contradictory, and it would have sent every existing working user off to restart for nothing.
2. The slug helpers are not guaranteed to be defined when `lib-memory-dir.sh` returns early, so the report printed an **empty slug as OK** and skipped the #144 check entirely. It now says it could not check, which is not the same as no problem — that distinction being the entire subject of this issue.

## Tests

`tests/test_capture_gap_notice.py` (8) and `tests/test_doctor.py` (5). Verified against the pre-fix scripts: the four asserting new behaviour fail without it. The rest are false-alarm guards protecting the new logic — one of which caught the mtime bug above.

Full suite: **880 passed, 41 skipped**.

## Not addressed here

`/reload-plugins` exists but the docs do not say whether it re-registers hooks. If it does, the remedy is a reload rather than a restart and the README should say so. Worth raising upstream either way — "enabling a plugin silently does nothing until restart" is arguably a Claude Code issue, not a plugin one.

@fmanimashaun offered to test on Windows/Git Bash, which we cannot do here — the scripts are written for bash 3.2 and MSYS, but that offer is worth taking up before release.


fdaviddpt:
## Review pass — four defects, all fixed in `2d717ec`

Ran four independent review agents over this branch. Each found something real, and each was the feature failing in the shape of the bug it was written to fix. Recording them because the PR description above was written before this pass and one of its claims was wrong.

**1. A pending notice could destroy the user's prompt.** The `systemMessage` path builds JSON with `jq`, and `jq` was the script's last command — so its exit status became the hook's. `jq` exits 2 on a usage error, and on `UserPromptSubmit` **exit 2 blocks the prompt and erases what was typed**. That would have fired on every prompt with a notice pending, silently, with `jq`'s stderr going to a log nobody reads in the moment. Proven end-to-end with a stub. The JSON is now built first and printed only if produced, falling back to plain text, and the hook ends with an explicit `exit 0`.

**2. The check could not catch the incident it was written for.** It was gated on a prior session-start stamp, so a fresh install wouldn't be warned about a session predating it. That sounds right and removes the entire point: during a mid-session enable *no hook runs*, so no stamp is written, so the one case this exists to report was the exact case it stayed silent for. It could only ever have caught a recurrence — and my own test locked that in. **The claim in the description above that "the signature is exact" was false as written.** The gate is gone; a fresh install sees the notice once per project and the wording covers both readings.

**3. The doctor misdiagnosed the failure it was built to name.** `capture-alive` was written *after* the hook's no-transcript early exit, so in a slug mismatch (#144) it was never written — and the verdict tested the generic "never fired" branch before the specific causes. So a slug mismatch **and** a missing Python both came out as "restart Claude Code", which fixes neither, on the one line `commands/doctor.md` told the operator not to second-guess. Now: a separate `post-tool-ran` marker answers "is the hook wired" before every early exit, the verdict names specific causes first, and the command doc asks for the `FAIL` lines alongside the verdict.

My comment claiming that marker was written "unconditionally" was false in exactly that case. Fixed, and the write now goes through a rename — the lesson `last-save.json` learned in #140.

**4. Test-harness honesty.** The fixture's `config.json` write is **inert**: `_LIB_MEMORY_DIR_LOADED=1` stops `lib-memory-dir.sh` before the config merge, so `features.recovery` reads true whatever the file says. The comment claimed it suppressed recovery. It doesn't — recovery is only absent because these fixtures don't create `last-save.json`, and anyone copying the pattern into a test that does would get a stray background fork. Said outright now. Transcript ordering also moved from a 50ms sleep to explicit spaced mtimes, since sub-second mtime resolution is the assumption that bit this feature once already.

### Known limits, now written down rather than discovered later

- `capture-alive` is per project, not per session — two concurrent Claude Code windows on one project can produce a spurious notice.
- A notice is consumed by the next prompt in *any* session, so a session closed before its first prompt hands it to the following one.

Neither is silent, and both are in the CHANGELOG.

Full suite: **882 passed, 41 skipped**.

Issue #243, opened 2026-07-30
NDC's commit ignores the `mv` result, so a failed truncate still rewrites the day stamp and logs a success


fdaviddpt:
## Code-read finding, `main` @ 2bb0d4f

`scripts/save-session.sh` ~603-629, inside the backgrounded NDC subshell, which runs under `set +e`:

```sh
if tail -c +$(( NDC_SRC_BYTES + 1 )) "$MEMORY_FILE" > "$NDC_TAIL" 2>/dev/null; then
    NDC_KEPT=$(wc -c < "$NDC_TAIL" | tr -d ' ')
    mv "$NDC_TAIL" "$MEMORY_FILE"
    ...
    if [ "$NDC_KEPT" -gt 0 ]; then
        printf '%s\n' "$(_remember_date +%Y-%m-%d)" > "$NOW_DAY_FILE" 2>/dev/null || true
    else
        rm -f "$NOW_DAY_FILE"
    fi
    [ "$NDC_KEPT" -gt 0 ] && log "ndc" "kept ${NDC_KEPT}b appended during compression"
```

**The `mv` result is never read.** `tail`'s is — carefully, with a whole `else` branch explaining why `now.md` is left untouched on failure (#173's defect (c)). The `mv` immediately after it has no such branch, and `set +e` means a failure does not stop the block either.

## What happens when it fails

`mv` can fail for the ordinary reasons: `ENOSPC` on the destination, a read-only or full volume, `EPERM`, an unlink race on the temp. When it does, `now.md` **still holds every byte it had**, including the span that was just appended to `today-*.md`. But the code proceeds as though the commit happened:

* **`NDC_KEPT == 0`** (the common case — nothing was appended during compression) → `rm -f "$NOW_DAY_FILE"`. The #141 day marker for content that is *still entirely in `now.md`* is deleted. The next round re-stamps that content with whatever day it runs on, so a session that crossed midnight files yesterday's entries under today.
* **`NDC_KEPT > 0`** → `NOW_DAY_FILE` is overwritten with today's date, for a `now.md` that still contains the *pre-compression* content the old stamp described. Same misfiling, one round earlier.
* Either way the log says **`kept ${NDC_KEPT}b appended during compression`**, or says nothing at all when `NDC_KEPT` is 0. Nothing anywhere records that the truncate did not happen.

[...]

Issue #242, opened 2026-07-30
NDC's commit `mv`s across filesystems, so the atomicity #142's design rests on is not the atomicity it gets


fdaviddpt:
## Code-read finding, `main` @ 2bb0d4f

`#142` replaced NDC's blind `: > now.md` with a partial truncate, and its own reasoning for why the commit is safe is atomicity:

> `mv`-over is atomic on the same filesystem

The shipped code does not put the temp file on the same filesystem. `scripts/save-session.sh:602`:

```sh
NDC_TAIL=$(mktemp "${TMPDIR:-/tmp}"/remember-ndc-tail-XXXXXX)
if tail -c +$(( NDC_SRC_BYTES + 1 )) "$MEMORY_FILE" > "$NDC_TAIL" 2>/dev/null; then
    NDC_KEPT=$(wc -c < "$NDC_TAIL" | tr -d ' ')
    mv "$NDC_TAIL" "$MEMORY_FILE"
```

`$MEMORY_FILE` is `${REMEMBER_DIR}/now.md`. `$TMPDIR` is a different filesystem in ordinary configurations — `/tmp` is `tmpfs` on most Linux distributions, and external storage mode (`data_dir`) can put `REMEMBER_DIR` on a different volume from `$TMPDIR` on any platform. Across filesystems `mv` is not a rename: it is copy-then-unlink, and `now.md` is *destroyed and rewritten in place* rather than swapped.

#142's suggested fix wrote `"$MEMORY_FILE.tmp"` — beside the target, where the atomicity argument holds. The implementation moved it to `$TMPDIR` and kept the argument.

## What it costs

The window is the copy, not a syscall. A machine that loses power, a `SIGKILL`, a full destination filesystem, or an `ENOSPC` partway through leaves `now.md` **truncated mid-copy**: not the old content, not the new content, a prefix of the tail. The save lock is held (#223/#224), so no concurrent writer is involved — this is single-process durability, which the lock cannot help with.

It also silently changes the file's inode, mode and — on a copy — its ownership semantics, which is why the `.remember/.gitignore` and permissions assumptions elsewhere are worth a look in the same pass.

## Fix

[...]

Issue #230, opened 2026-07-30
post-tool-hook pays the same resolution prefix on every tool call


fdaviddpt:
Split out of #227 / #229, which fixed the prompt hook only.

`post-tool-hook.sh` sources `resolve-paths.sh` → `detect-tools.sh` → `bootstrap-dirs.sh` → `log.sh` on **every tool call**. @MargeryOlethea measured the shared prefix at a median of 3401-3784ms on Windows 11 ARM64 under QEMU, and `detect-tools.sh` alone at a stable 1.0-1.3s across all 14 runs, independent of `jq`. They flagged the per-call figure as *not* a claim — successful silent `PostToolUse` runs are not persisted to transcripts, and their manual invocation (6.5s / 11.6s / 31.9s) almost certainly included a real `save-session.sh` fork past its cooldown. So step one here is our own measurement, not their number.

#229 gave `user-prompt-hook.sh` a fast path by replaying an already-published resolution (`scripts/lib-env-cache.sh`). It cannot serve `post-tool-hook.sh` as-is, because that hook needs `config()` for cooldowns and thresholds — which means publishing the **merged config file itself**, at a stable path instead of today's `remember-config-$$.json`, with its own lifetime, trap and cleanup semantics (`tests/test_tmpdir_cleanup.py`). That is where the risk is, and it is why it was left out.

Two behaviour-preserving wins that need none of that, and should come first because they also speed up the cold path everywhere:

- **`_resolve_memory_project_dir` forks `git rev-parse --path-format=absolute` (~200ms each on that box) to detect a linked worktree.** A linked worktree always has `.git` as a *file*; a main checkout has it as a *directory*. So `[ -d "$proj/.git" ]` — a shell builtin — proves "no redirect needed" outright for the common case, with everything else falling through to today's `git` call unchanged. Careful: the *absence* of `.git` proves nothing (a subdirectory of a worktree still needs the rev-parse), so this is a short-circuit on the positive answer only. `tests/test_worktree_memory.py` is the pin.
- **`bootstrap-dirs.sh` and `log.sh` run `mkdir -p` unconditionally**, twice per invocation, for directories that exist on every run after the first. Gate on `[ -d ]`.

Then measure again before touching the merged config. Method that worked in #229: a PATH shim per external command that appends its name to a log and `exec`s the real binary, asserted as a spawn budget in a test — count executions, not milliseconds, because wall clock is what differs between platforms and spawn count is what causes it.

[...]

Issue #227, opened 2026-07-30
`user-prompt-hook` spawns 27 processes to print a timestamp — 8.7s p50 blocking every prompt on Windows


MargeryOlethea:
## Symptom

Every prompt submission blocks for a measured **p50 of 8.7s, max 30.0s**, with 6 hard timeouts, on Windows 11 ARM64 (Git Bash, under QEMU). Aggregated from 256 `UserPromptSubmit` hook records across 50 sessions over 5.7 days:

| hook | runs | p50 | max | timed out |
|---|---|---|---|---|
| `UserPromptSubmit` | 256 | **8718ms** | 30003ms | 6 |

That is ~37 minutes of blocking wait across those 256 prompts. Claude Code's guidance is >2s is a slow hook for per-prompt events, so this is ~4x over.

## Root cause

`scripts/user-prompt-hook.sh` sources three libraries to print a timestamp:

```bash
REMEMBER_PATHS_SOFT_FAIL=1 source "$(dirname "$0")/resolve-paths.sh" || exit 0
source "$(dirname "$0")/bootstrap-dirs.sh"
source "$(dirname "$0")/log.sh" 2>/dev/null
```

Traced with `bash -x`, one invocation of the hook makes **27 external process spawns**. Cumulative cost, isolated by sourcing each library in turn:

| sourced through | cumulative |
|---|---|
| `resolve-paths.sh` | 1862ms |
| `+ bootstrap-dirs.sh` | 8372ms |
| `+ log.sh` | 12584ms |

`bootstrap-dirs.sh` is the biggest single contributor (~6.5s), almost entirely via `lib-memory-dir.sh`, which on every prompt runs `git -C "$proj" rev-parse --path-format=absolute ...` plus several `jq` config merges to re-resolve `REMEMBER_DIR`.

Measured per-spawn cost on this platform, which is what turns a benign design into a blocking one:

```
git rev-parse --path-format=absolute --git-common-dir   194ms / 190ms / 234ms
jq (lookup)                                             ~200ms
whoami                                                   716ms
date                                                     813ms
```

On a normal Linux box 27 spawns is ~300ms and nobody notices. At 150-800ms per spawn it is ~8s. The design is fine; the platform assumption underneath it isn't.

[...]

PR #244, opened 2026-07-30
feat(#226): record save.lock hold times, so the 30s timeout can be set from data instead of intuition

fdaviddpt:
Part of #226 — **not** `Closes`. The instrument ships; the measurement has not been taken, and closing this would bury exactly the question the issue exists to ask.

## What this is, and what it deliberately is not

#226 says what would settle it: *"instrument `lock_acquire` / `lock_release` in `save-session.sh` across a normal working day and record the held-duration distribution. Then set the default from the tail of that distribution rather than from intuition."*

I cannot do the second half. There is no working day of real saves here, and a synthetic run would produce a number that **looks measured and is not** — which is precisely the defect this issue complains about, reproduced one layer up. So **`REMEMBER_NDC_COMMIT_LOCK_TIMEOUT` is untouched at 30s.** What ships is the thing that makes the number obtainable on a real machine:

```bash
export REMEMBER_LOCK_TIMING=1   # in the shell Claude Code launches hooks from
# ...work normally for a day...
scripts/lock-timing-report.sh
```

```
lock-timing: ok  file=~/.remember/<slug>/logs/lock-timing.tsv  records=418

lock            prec     n  held_p50  held_p90  held_p99  held_max  wait_p50  wait_p90  wait_p99  wait_max timeouts
save.lock         us   197      4210      9840     21030     24118         0         1      2004     30001        1
staging.lock      us   210        31        44        88       201         0         0         1        12        0
```

`timeouts` is the column that answers #226 outright: for `save.lock` each one is an NDC commit that lost its wait, skipped, and duplicated a span into `today-*.md` — the outcome the bounded wait was chosen to avoid, arrived at anyway. Today that event leaves a log line and nothing a distribution can be built from.

The recorder lives in `lib-lock.sh` rather than in `save-session.sh`, so `staging.lock` and `consolidation.lock` come along for free and the comparison in the table above is possible at all. #234's staging figures (~30ms/append, ~200ms for a 3-file retire) become reproducible rather than a thing someone did once by hand with a 1ms poller.

### 1. Opt-in, not always-on

`save-session.sh` runs on a `PostToolUse` hook. #227 measured that path at a p50 of 8.7s on Windows/ARM64, #230 split out the prefix paid per tool call, and #204 is an external report of the plugin blocking a user's prompts. So the cost is not abstract.

**Always-on:** one `wc` per record for the cap check on every platform, plus **two `date` spawns per lock use** on bash < 5 — every save, every machine, forever, to answer a question asked once. It would also bias its own numbers by slowing the path it measures.

**Env-gated:** on the disabled path `lock_acquire` and `lock_release` do one string comparison (`[ "$_LOCK_TIMING" != 1 ]`) and delegate to the untouched primitive. **Zero spawns, zero files.** That is a pinned property, not a claim: `test_with_the_recorder_off_nothing_is_written_and_nothing_extra_is_spawned` counts executions through the `tests/spawn_counting.py` PATH shims and asserts `off < on` — make the recorder unconditional and it goes red.

The trade I am accepting: nobody's machine records anything unless they choose to. Given the alternative is a measured slowdown in the hook path a user has already filed a bug about, that is the right way round.

### 2. Where the record goes, and what bounds it

`$REMEMBER_DIR/logs/lock-timing.tsv`, overridable with `REMEMBER_LOCK_TIMING_FILE`. TSV, one row per lock use:

```
# ts_ms  lock  event  outcome  wait_ms  held_ms  precision  pid
```

Bounded at `REMEMBER_LOCK_TIMING_MAX` **lines** (default 5000, ~350KB). This repo has just spent a day on unbounded things — #605's 1.6 MB event, its 770 MB line buffer — and an instrument that grows forever on someone's machine would be a poor way to end it.

**It stops; it does not roll.** At the cap, recording halts, a `# CAPPED` line goes *into the file*, and a `.capped` sibling marker makes the stop a builtin `[ -e ]` test rather than a re-scan on every later record. Rolling was the obvious choice and is the wrong one here: the file's whole purpose is the **tail** of the distribution, and a rolling window silently discards the oldest records while presenting the remainder as complete. The report surfaces the cap and says to read `held_max` as a floor rather than as the maximum.

`pid` is `${BASHPID:-$$}`, not `$$` — the NDC commit runs in a backgrounded *subshell* of `save-session.sh`, where `$$` is identical to the parent's, so the two sides of the very contention #226 is about would otherwise be indistinguishable in the file.

### 3. What reads it

`scripts/lock-timing-report.sh`. A distribution nobody can see is not an instrument, and a README one-liner would rot silently — this one is exercised by two tests. Nearest-rank percentiles, insertion-sorted in `awk` (no `asort`; that is gawk-only and macOS ships the one-true-awk). README documents both the env var and how to read the output.

### 4. The 30s default is not touched

It is the *output* of the measurement this enables, not part of it. I do not think the evidence justifies a change yet, and I would rather it move on a number than on a second intuition. The `#226` comment noting staging's 10s came from data is exactly the standard `save.lock` should be held to.

## Resolution is disclosed, not assumed

There is no portable spawn-free millisecond clock. Three tiers, probed once, and **the winner is written into every row**:

| tier | source | cost | where |
|---|---|---|---|
| `us` | `EPOCHREALTIME` | 0 spawns | bash ≥ 5 |
| `ms` | `date +%s%N` | 1 spawn | GNU coreutils |
| `s` | `date +%s` | 1 spawn | macOS `/bin/bash` 3.2 + BSD `date` |

A `python3` fallback was considered and **rejected**: at ~60ms per spawn it costs more than a whole `staging.lock` hold, so it would advertise millisecond precision while inventing most of the milliseconds. Reading a number at a finer resolution than it was taken at is the same false confidence #226 was filed about, so the coarse answer is given honestly instead. One second is useless for `staging.lock` and adequate for `save.lock`, whose holds are the seconds-long ones this issue is asking about.

The `us` tier also tests `BASH_VERSINFO[0] -ge 5`, not just the variable: `EPOCHREALTIME` is an ordinary variable on bash 3.2, so an exported stale value would pin every row to one frozen timestamp and report **every hold as 0ms** — a plausible number, wrong, and indistinguishable from a very fast machine.

## The trap: not trading a loud failure for a quiet one

A measurement that silently did not happen reads identically to one that showed nothing. So:

* An unwritable destination, or `REMEMBER_LOCK_TIMING=1` with no `REMEMBER_DIR` and no explicit file, is **announced once per process** — through `log` when the caller sourced `log.sh`, otherwise on stderr — and locking carries on regardless. Locking must not fail because its instrument did; it must also not fail silently.
* Every append is braced (`{ printf … >> "$f"; } 2>/dev/null`) so a redirect failure is caught inside the suppression rather than escaping it — the #204 lesson, reused.
* The report has three states, following `tests/pep604_floor.py`'s contract: `ok`, `CAPPED`, and **`skipped` at exit 2 with the reason and how to fix it**. A report that printed an empty table on a file that was never written would read exactly like one taken on an idle machine, and those are the two answers most worth telling apart.
* A `release` with no matching `acquire` in the same process is recorded as `unpaired` with `-` for the duration, not as a duration computed from a start that does not exist.

Two `set -e` hazards were removed while writing it: `[ … ] && return 0` as a statement's last command is the statement's exit status, and `save-session.sh` runs under `errexit`. The existing instances in `lib-lock.sh` are safe only because those functions are called in condition position; these ones would not have been.

## Red, then green

Tests first. Red on `main` with only the test file present — 6/6, and each one fails for the right reason:

```
FAILED test_a_real_hold_is_recorded_with_a_plausible_duration
  AssertionError: REMEMBER_LOCK_TIMING=1 but nothing was written to …/lock-timing.tsv —
  a measurement that silently did not happen reads identically to one that showed nothing
FAILED test_a_lost_wait_is_recorded_as_a_timeout_carrying_the_wait_it_spent
FAILED test_with_the_recorder_off_nothing_is_written_and_nothing_extra_is_spawned
FAILED test_the_cap_stops_and_discloses_rather_than_rolling
FAILED test_the_report_turns_the_file_into_a_distribution
FAILED test_the_report_says_skipped_rather_than_reporting_its_own_silence
6 failed in 16.09s
```

Green, with the three clock-conversion cases added:

```
9 passed in 17.34s
```

**"Would this still pass if the recorder did nothing?" — no, and not only because of the file check.** The hold test takes a real 2.5s hold and a real immediate one and requires them to be **≥ 1000ms apart**, which kills a recorder that writes a constant, and bounds each one, which kills a recorder that writes seconds or nanoseconds into a millisecond column. 2.5s rather than something tighter because the assertion has to hold at *one-second* resolution, where the pair records as 2000/3000 against 0/1000.

The clock conversion is driven directly with an injected reading, including the `LC_NUMERIC=de_DE` comma separator and a truncation case, because **no single machine reaches all three tiers** — this Mac's `/bin/bash` 3.2 only ever takes `s`, a bash-5 Linux runner only ever `us`. Without it two of the three ship untested on whichever runner does not happen to have them.

Verified by hand on bash 3.2 (`s` tier, so the report shows a 0.3s hold as `0` and a 1.2s hold as `1000` — honest, and labelled `prec=s`).

Full suite: **1096 passed, 42 skipped**, coverage 94.13%.

## What else the `save.lock` path does wrong

Asked explicitly rather than assumed answered, and it found two, both filed rather than fixed here — each wants its own red test:

* **#242 — the commit `mv`s across filesystems.** #142's design rests on *"`mv`-over is atomic on the same filesystem"*, and its own suggested fix wrote `"$MEMORY_FILE.tmp"`. The shipped code uses `mktemp "${TMPDIR:-/tmp}"/…` and keeps the argument. `/tmp` is tmpfs on most Linux distributions, and external `data_dir` mode can split them on any platform — so the `mv` is copy-then-unlink, `now.md` is destroyed and rewritten in place, and a crash mid-copy leaves it truncated. The save lock is held throughout, which is exactly why the lock cannot help: this is single-process durability.
* **#243 — the commit ignores the `mv` result.** `tail`'s result is checked with a whole `else` branch explaining why `now.md` is left untouched (#173's defect (c)); the `mv` on the next line has none, under `set +e`. On failure `now.md` still holds every byte — but the code deletes or rewrites the `#141` day marker anyway, and logs `kept ${NDC_KEPT}b` or nothing at all. Memory is not lost, it is **misfiled**, and the log reads like a clean run: #142/#225's signature, in the one file that has been hardened against it three times.

## Docs

`CHANGELOG.md`, and `README.md` for `REMEMBER_LOCK_TIMING` / `_FILE` / `_MAX` plus a **Measuring lock hold times** section with the worked example above — an instrument nobody can find is not an instrument.


fdaviddpt:
## The test was wrong, not the conversion — and the CI output proves it rather than my reasoning about it

Pushed `4bc7f7e`.

**`ts_ms` is correct on Linux.** Two independent pieces of evidence, both from the failing run:

1. `tests/test_lock_timing.py ......FFF` — the **first six passed on ubuntu**, including `test_a_real_hold_is_recorded_with_a_plausible_duration`, which runs the real recorder at the `us` tier on bash 5 and requires a 2.5s hold to land in [1000, 15000]ms and sit ≥1000ms above an immediate one. A conversion producing wrong milliseconds on Linux fails that test; it passed on all four legs.
2. The three failure values are themselves the proof. `1785437689796` / `…825` / `…843` are well-formed 13-digit epoch-**milliseconds**, incrementing 20-50ms across three subprocess launches. The failing test *forces* `_LOCK_TIMING_PRECISION="us"`, so the `us` branch definitely executed — it read something the harness did not put there, and converted it correctly. Broken arithmetic produces garbage, not a valid current timestamp.

What no-opped was the injection: on bash ≥ 5 `EPOCHREALTIME` is generated on every reference, so assigning it never reaches the code under test. On macOS's bash 3.2 it is an ordinary variable, so it did — green there, red on Linux.

## Why that is worth more than a red CI leg

The mechanism was **green while asserting nothing**, on exactly the platform whose tier it claimed to cover. Had macOS been the only leg, this would have shipped as coverage that wasn't. That is the same defect class as the rest of this PR — a measurement that silently did not happen reading identically to one that showed something — arriving inside the test written to prevent it.

## The fix, with the assertions unchanged

The three conversions are now functions of their argument (`_lock_timing_us_to_ms`, `_lock_timing_ns_to_ms`, `_lock_timing_s_to_ms`), and `_lock_timing_now` calls them with the reading it took. Same code the live caller runs — not a parallel copy — and drivable on every platform, because passing an argument cannot no-op.

Nothing was relaxed. The cases are **stronger**: the original three, plus a fraction-less reading, a malformed reading yielding `0` rather than a shell arithmetic error (a clock that cannot be read must not take locking down with it, and must not produce a *plausible* number either), and — new — the **`ms` tier, which no runner in this matrix ever executes live**, including the literal-`N` string BSD `date +%s%N` prints.

Teeth verified by mutation: dropping the sub-second term from `_lock_timing_us_to_ms` turns 2 of the 9 cases red.

Local full suite: **1102 passed, 42 skipped**, coverage 94.13%.

Noted on both points — I will report findings and leave the filing to you from here.

PR #323, opened 2026-08-08
fix(save-session): never let compression drop a blocked/pending status

turbomotioncat:
## Problem

`prompts/save-session.prompt.txt` compresses each save into exactly one sentence, but has no rule protecting "blocked / pending / not-yet-live" status from being treated as droppable filler alongside things like "successfully" or "in order to".

Real-world hit: a session deployed app code but was explicitly blocked on a manual step (a human running a data upload with credentials the assistant doesn't have access to). The assistant told the user clearly, twice, in chat. The saved `.remember/now.md` entry kept "...deployed" and silently dropped the entire "blocked on you" caveat. The user read the memory entry later, believed the feature was fully live, and only found out otherwise by checking the app directly.

## Fix

Two small, coordinated edits to `prompts/save-session.prompt.txt`, in its existing style:

- The format bracket now explicitly calls out appending blocked/incomplete/waiting-on-a-manual-step status as a short clause in the same sentence.
- A new rule bullet (next to the existing non-destructive-compression rule) states this plainly: blocking/pending/not-yet-live status is a fact, not filler, and must survive compression even if the sentence gets longer.

No changes to the `## {{TIME}} | {{BRANCH}}` header contract, the four `{{PLACEHOLDER}}`s, or the SKIP logic.

## Testing

- `tests/test_prompts.py` — all 10 tests pass unmodified (`python3 -m pytest tests/test_prompts.py -v --no-cov`), including `test_build_save_prompt_with_real_templates`, which loads this exact file.
- `scripts/run-tests.sh` section 6 ("Prompt templates") passes — file exists, `{{TIME}}` placeholder intact.
- Manually built a synthetic prompt via `pipeline.prompts.build_save_prompt()` against a "deployed but blocked on a manual step" transcript and confirmed the new rule text assembles correctly into the final prompt sent to the summarizer.

### DIFF
--- a/prompts/save-session.prompt.txt
+++ b/prompts/save-session.prompt.txt
@@ -3,12 +3,13 @@ You are summarizing a Claude Code session for a daily memory log.
 Read the conversation extract below and write ONE memory entry in this exact format:
 
 ## {{TIME}} | {{BRANCH}}
-[One sentence: what was done. Be specific — mention files, MR numbers, issue numbers.]
+[One sentence: what was done. Be specific — mention files, MR numbers, issue numbers. If the work is blocked, incomplete, or waiting on a manual/human step (e.g., missing credentials, a manual upload, human approval), append that status as a short clause in the same sentence — never omit it.]
 
 Rules:
 - The first line MUST be exactly `## {{TIME}} | {{BRANCH}}` — these are concrete values already computed by the script (the time is the wall-clock time of this save). Copy them verbatim. Do NOT invent your own header (e.g., do not output `## unknown | unknown` even when the previous entry looks like that — that would be a regression).
 - ONE sentence only. Short and specific.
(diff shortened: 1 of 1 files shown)


fdaviddpt:
Merged — thank you.

What made this easy to accept was the framing: you didn't report "the summary was bad", you named the exact class of fact that compression was allowed to eat. "Blocking status is a fact, not filler" is the right line, and it sits properly next to the non-destructive-compression rule rather than fighting it. The failure you described — reading `...deployed` later and believing it was live — is the worst outcome this plugin can produce, because a memory entry that is wrong is worse than no memory entry at all.

Two things I checked before merging, in case they're useful to you: `test_build_save_prompt_with_real_templates` does load this exact file, so the 10 green tests are real coverage of the assembly path, and the header contract and the four placeholders are untouched. Shipping in the next release.

If you use Claude Code enough to have hit this, you may also get some mileage out of our other one — [claude-supertool](https://github.com/Digital-Process-Tools/claude-supertool). Same idea applied to tool calls instead of memory: batch reads/greps/edits into a single call, with post-edit validation and rollback. It's stdlib-only Python and source-available like this one. No obligation, just adjacent.

Issue #324, opened 2026-08-08
#258's regression test writes the cooldown marker to a path the hook no longer reads, so every corrupt value passes trivially


jmossie82:
### Summary

`test_a_corrupt_cooldown_marker_does_not_kill_the_hook` (`tests/test_git_backup_silent_stops_257.py`) writes the marker to a path the hook no longer reads, so every corrupt value passes trivially and **#258's regression is currently unguarded**.

### The mismatch

- Test writes `<store>/.last-git-backup-ts` (`:433`).
- Hook resolves the marker through `_gb_common_dir` and reads `$GB_STATE_DIR/last-git-backup-ts`, i.e. `<store>/.git/remember/last-git-backup-ts` (`50-git-backup.sh:125`, `:155`).

Confirmed by running a backup and globbing for the file rather than by reading the code:

```
marker files created:
    .git/remember/last-git-backup-ts = 1786175184
legacy dotted path exists? False
```

The cleanup helper at the top of the same file already globs **both** names (`:458-459`), which looks like the trace of the path having moved.

### Consequence

With the marker written where the hook reads it, `08` reaches the arithmetic and bash reports `line 172: 08: value too great for base` — the octal case a digits-only guard cannot catch. Written to the dotted path, the cooldown block is never entered at all, so the test asserts nothing about the guard it exists to protect.

### Fixing the path is necessary but not sufficient

Measured on `/bin/bash` 3.2.57: with `08` at the real path the hook prints the octal diagnostic but still returns **rc=0** and commits, because the failing arithmetic abandons the whole `if/then` body and `$ELAPSED` is never referenced. So the existing assertions (`proc.returncode == 0`, `_slug_is_committed`) stay green on the macOS leg while the same input behaves differently on newer bash.

Asserting on `proc.stderr` makes it fail on any version:

```python
for _phrase in ("syntax error", "invalid arithmetic operator",
                "value too great for base"):
    assert _phrase not in proc.stderr
```

With the path fixed and that assertion added, `08`/`09` fail against the current hook and pass once `10#` is added after the existing `case` guard.

### Scope

Filing this separately from #322 on purpose: it is a test-coverage bug in `50-git-backup.sh`'s suite, independent of the `save-session.sh` reads that issue is about. I have the patch and can open a PR if useful — say which you'd prefer.

Issue #277, opened 2026-07-31
`dispatch` discards a dying hook's stderr and reports it as a nameless `hook failed`, so a fatal hook error reaches nobody


fdaviddpt:
Found while fixing #258. Not folded into that PR: the line lives in `scripts/log.sh`, which is being worked concurrently, and this is a different defect from the four in that PR.

## The line

`scripts/log.sh`, in `dispatch()`:

```sh
REMEMBER_PROJECT="${PROJECT_DIR:-.}" "$hook" 2>/dev/null \
    || log "dispatch" "hook failed: $event/$(basename "$hook")"
```

A hook that dies loses **both** halves of its report at once: `2>/dev/null` throws away what it said, and the fallback line names the file but not the reason or the exit status. The result is a log entry that proves something is broken and cannot say what, which is the least actionable form a report can take.

## Why it is not merely untidy

It is the swallow that made #258's worst case invisible, and that case is fatal rather than quiet. `50-git-backup.sh` runs `set -u`, and its cooldown marker's contents went straight into `$(( ))`. Bash evaluates file content there as an *arithmetic expression*, so a marker holding a letter is an unbound-variable reference and the hook dies on that line — above the lock, above the `add`, above the commit:

```
== garbage (content=[garbage-text])
   rc=127 out=[bash: garbage: unbound variable]
== nul-ish (content=[12x34])
   rc=1 out=[bash: 12x34: value too great for base (error token is "12x34")]
```

Both messages name the problem exactly. Neither reaches any log, any file, or any human. What the user gets instead is `hook failed: after_save/50-git-backup.sh`, with no cause and no exit status, in the log #252 demonstrated nobody reads — and their memory silently stops being backed up from that moment on.

The marker guard is fixed in the PR for #257/#258/#260/#261, so that specific route is closed. The swallow is not, and it applies to **every** hook on every dispatch point: the next `set -u` slip, the next missing binary, the next `E2BIG`, all arrive as the same causeless line.

## The shape

The house defect, one layer up and in the reporting path itself: three states collapsed into two. There is `ran and succeeded`, `ran and failed *for this reason, with this status*`, and today the middle one is spelled the same as a shrug. `_push_and_report` in `50-git-backup.sh` is the in-repo example of getting this right — pushed / deferred-will-retry / rejected-stopped.

## Suggestions, not a specification

Whoever picks this up should design it; some constraints that are real:

[...]

Issue #258, opened 2026-07-31
Cooldown and lock robustness in 50-git-backup.sh: a corrupt marker bypasses the throttle, and the marker is only written on success


fdaviddpt:
Three robustness defects in `hooks.d/after_save/50-git-backup.sh`, all around the cooldown and lock. None is as serious as #253 or the silent stops filed alongside this; they are grouped because they share a file and a shape.

## 1. A corrupt cooldown marker silently bypasses the cooldown (`:88-90`)

```sh
    LAST_MOD=$(cat "$COOLDOWN_MARKER" 2>/dev/null || echo 0)
    ELAPSED=$(( $(date +%s) - LAST_MOD ))
    if [ "$ELAPSED" -lt "$BACKUP_COOLDOWN" ]; then
```

`LAST_MOD` is unvalidated file content fed straight into `$(( ))`. A truncated or garbage marker — a crash mid-write, a partially synced file — makes the arithmetic error out, leaves `ELAPSED` empty, and then `[ "" -lt 900 ]` errors too. The cooldown is skipped, and the full add/rm/commit runs on **every save** from then on, with nothing saying why.

The fix is the same two-line numeric guard #253's PR already had to add for its rejection counter:

```sh
case "$LAST_MOD" in ''|*[!0-9]*) LAST_MOD=0 ;; esac
```

## 2. The cooldown marker is written only when the commit succeeds

So on a store where commits fail — see the sibling issue about `:201` — the throttle never engages at all. The two defects compound: the store stops backing up *and* starts doing the full work on every single save, which is the opposite of what a cooldown is for. Neither is visible.

## 3. A narrow re-entrancy window on the flock path

The subshell's `trap 'rm -f "$LOCK_FILE"' EXIT` unlinks the lock file. If a second instance opens the file just before that `rm`, it locks an unlinked inode while a third instance creates a fresh one — two concurrent backups on the same store.

**Lower confidence than the other two, and worth re-deriving before acting.** The agent that found it checked the part that is *fine* — the lock lives on fd 9 and is correctly inherited by the background subshell — and argues the `rm` is simply unnecessary on the flock path. That is a small change, but concurrency changes deserve their own reproduction rather than a reading.

## Provenance

All three found by the agent implementing #253's part 1 and reported rather than folded in. Items 1 and 2 verified independently against `main` (`:88-90`); item 3 is recorded as unverified, deliberately.

Issue #79, opened 2026-05-25
Windows: re-enable shell-subprocess tests via Git Bash on PATH


fdaviddpt:
## Context

In #78 we marked several test modules `skipif(win32)` to get the new 3-OS CI matrix green:

- `tests/test_log_sh.py` — bash dispatch + ownership checks
- `tests/test_migration.py` — bash bootstrap-dirs.sh
- `tests/test_security_fixes.py` — RCE prevention in `safe_eval`, path-with-space cleanup, `_jq_fallback` quoting
- `tests/test_path_resolution.py` — POSIX path layouts + bash subprocess capture

Three of these (`test_log_sh`, `test_migration`, `test_security_fixes`) test logic that **does run on Windows** when Claude Code invokes hooks via Git Bash. Skipping them means we ship no Windows security/regression coverage for those code paths.

## Root cause of the skip

The GitHub Actions Windows runner's `bash` on `PATH` resolves to the WSL launcher (`C:\Windows\System32\bash.exe`), not Git Bash. WSL isn't installed on the runner, so every `subprocess.run(["bash", "-c", script], ...)` emits a UTF-16-encoded "no distributions installed" error and returns exit 1. The tests see garbage output and fail.

## Proposed fix

In `.github/workflows/tests.yml`, on the Windows row, prepend Git Bash to `PATH` so `subprocess.run(["bash", ...])` resolves to `C:\Program Files\Git\bin\bash.exe`:

```yaml
- name: Use Git Bash on Windows
  if: runner.os == 'Windows'
  shell: pwsh
  run: echo "C:\Program Files\Git\bin" | Out-File -FilePath $env:GITHUB_PATH -Append -Encoding utf8
```

Then revert the `pytestmark = pytest.mark.skipif(sys.platform == "win32", ...)` blocks added in #78 on:

- `tests/test_log_sh.py`
- `tests/test_migration.py`
- `tests/test_security_fixes.py`

Leave `test_umask.py` skipped — POSIX mode bits genuinely can't exist on NTFS. Possibly `test_path_resolution.py` too — `/c/Users` vs `C:\Users` path-form assumptions may need targeted per-test fixes.

## Out of scope

- Per-test debugging once Git Bash runs — may surface real Windows bugs in the scripts that need actual fixes (path normalization, line endings, etc.).
- `test_umask.py` — NTFS lacks Unix permission bits, not a CI configuration problem.

## Test plan

- [ ] Workflow `PATH` shim added.
[...]

PR #325, opened 2026-08-08
fix(#322): save-session.sh's two cooldown markers reach $(( )) unvalidated, and a digits-only guard still lets octal through

### LINKED ISSUE #322: save-session.sh's two cooldown markers reach $(( )) unvalidated — one skipped cooldown per corruption, and a digits-only guard still lets octal through
> [!IMPORTANT]
> **The mechanism, settled — this supersedes both the original "Impact" section below and my first correction comment.**
>
> Bash abandons **the entire `if … then … fi` body** at the failing arithmetic and resumes after `fi`. Measured on `/bin/bash` 3.2.57 and 5.2.37. Three consequences, all of which I got wrong earlier:
>
> 1. **Nothing dies.** Not in `save-session.sh`, and not in `50-git-backup.sh` either — its `$ELAPSED` sits inside the abandoned body, so `set -u` never sees an unbound variable. Against the real hook: `rc=0`, backup proceeds, commit lands. My earlier "#258's death mode" framing was wrong.
> 2. **There is exactly one diagnostic**, not two. `[ "$ELAPSED" -lt … ]` is never reached, so `integer expression expected` is never emitted.
> 3. **Both markers self-heal.** `save-session.sh:158` rewrites the save marker *after* `fi`, and `:585` rewrites the NDC marker on precisely the path a corrupt one allows. So one corruption costs **one skipped cooldown and one stderr line**, then repairs itself — not a throttle stuck off.
>
> The unguarded reads are still real, and the `case` guard is still the right fix — but the severity is modest, and I withdraw the "17 occurrences" observation from the body below as evidence: with both gates self-healing, that count needs an external re-corrupting writer I never identified.
>
> Still standing and independently verified: `08`/`09` pass a digits-only guard and then fail as octal (so `10#` is needed alongside `case`), and ShellCheck does not detect this class even with `-o all`.

---

### Summary

`#258` hardened the cooldown marker read in `50-git-backup.sh`, and the CHANGELOG describes the guard as *"A numeric guard now falls back to `0`, which self-heals, because the run it allows is the run that rewrites the marker."*

The identical unguarded pattern is still present in `scripts/save-session.sh`, reading the NDC cooldown marker `tmp/last-ndc.ts`. It bit me on 0.15.0 — 17 times in one project, and a second project on this machine still had a corrupt marker waiting to trigger it.

### The code (0.15.0, `scripts/save-session.sh:564-568`)

```bash
if [ "$RUN_NDC" = true ] && [ -f "$NDC_MARKER" ]; then
    NDC_MOD=$(cat "$NDC_MARKER" 2>/dev/null || echo 0)
    NDC_COOLDOWN=$(config ".cooldowns.ndc_seconds" 3600)
    [ $(( $(date +%s) - NDC_MOD )) -lt "$NDC_COOLDOWN" ] && RUN_NDC=false   # line 567
fi
```

`NDC_MOD` goes into `$(( ))` bare, so bash evaluates its **value** as an arithmetic expression. A stray character is therefore a syntax error, not a bad number.

### Observed

`.remember/logs/hook-errors.log`, 17 lines, all identical in shape:

```
/Users/…/remember/0.15.0/scripts/save-session.sh: line 567: 1786158837;: syntax error: invalid arithmetic operator (error token is ";")
```

The marker held `1786158837;` — a valid epoch with a trailing `;`.

### Reproduction

```bash
$ NDC_MOD="1786158837;"; echo $(( $(date +%s) - NDC_MOD ))
bash: 1786158837;: syntax error: invalid arithmetic operator (error token is ";")
```

### Impact — worse than the throttle being bypassed

The intuitive reading is that the `[ ... ]` test fails, so `RUN_NDC` stays `true` and compression merely runs too often. That is backwards, exactly as `#258` itself notes for the git-backup case.

An arithmetic syntax error is **fatal to the shell**, so the script dies at line 567. Verified on `/bin/bash` 3.2.57, **with and without `set -e`** — the statement after the gate never runs, exit 1:

```bash
$ /bin/bash -c 'set -e; BAD="1;"; [ $(( 10 - BAD )) -lt 5 ] && F=false; echo "AFTER_GATE_RAN"'
/bin/bash: 1;: syntax error: invalid arithmetic operator (error token is ";")
$ echo $?
1
```

(`#258`'s note about the shape differing between bash 3.2 and bash 5 concerned `set -u` in `50-git-backup.sh`. `save-session.sh` sets `-e` at line 56, and here the abort is at the line itself. I only have bash 3.2.57 available, so I have not confirmed the bash 5 shape.)

So on every affected save:

- **NDC compression never runs at all** — the gate is inside Step 8, and the entire `if [ "$RUN_NDC" = true ]` body below it is skipped.
- The **"remove empty autonomous logs" housekeeping** at the foot of the file is skipped too.

### Why it is effectively invisible

- The gate sits in **Step 8 of 14**. Step 7 has already appended the entry and saved the position, so the save genuinely succeeds.
- `scripts/doctor.sh` therefore reports `VERDICT: capture is working` with a recent successful save — which is true, and hides that the back half of the script has not executed for weeks.
- `trap cleanup EXIT` (line 102) still releases the lock, so nothing jams and no subsequent save is blocked.

The only trace anywhere is `hook-errors.log`, which `doctor` surfaces as a generic `WARN … is non-empty`. Same class as `#258`: *"Memory destroyed, log reads clean."*

### It is not a one-off

Scanning every `.remember/tmp/last-ndc.ts` on this machine found **two** corrupt markers across seven projects — a second project's marker held `1777939222;` and had been sitting there since May, guaranteed to abort its next save.

Because the marker is only rewritten *after* the gate is passed, a corrupt value is permanent — the same "stayed corrupt because it is only written on success" property `#258` identified.

### Provenance: unknown

I could not determine how the `;` got written. Nothing in 0.15.0 or 0.5.0 can produce it — both write the marker with a bare `date +%s > "$NDC_MARKER"`.

I did test `#258`'s "partially synced file" hypothesis on this machine and **ruled it out**: macOS Desktop & Documents iCloud sync is off (`FXICloudDriveDesktop = 0`) and `~/Documents` is a real local directory, not a redirect. A stray `;` also is not a plausible truncation artifact of a digit stream.

Which is rather the point of the guard — the reader should not be load-bearing on the writer being perfect.

### Suggested fix

The same shape as `#258`'s, applied to this reader:

```bash
if [ "$RUN_NDC" = true ] && [ -f "$NDC_MARKER" ]; then
    # Braces so the redirect's own failure is silenced too, not just tr's: if the
    # marker is unlinked between the -f test and this read, the "No such file"
    # comes from the shell.
    NDC_MOD=$({ tr -cd '0-9' < "$NDC_MARKER"; } 2>/dev/null)
    [ -n "$NDC_MOD" ] || NDC_MOD=0
    NDC_COOLDOWN=$(config ".cooldowns.ndc_seconds" 3600)
    [ $(( $(date +%s) - NDC_MOD )) -lt "$NDC_COOLDOWN" ] && RUN_NDC=false
fi
```

Falling back to `0` self-heals for the reason `#258` gives: the run it permits is the run that rewrites the marker. It also preserves existing behaviour — `0` is what the old `|| echo 0` already meant for an unreadable file.

Verified locally against the patched script: `bash -n` clean, and `tests/test_ndc_day_boundary.py` + `tests/test_save_session_gates.py` pass 31/31 (the `2**40` sentinel those tests use to suppress NDC still parses and still suppresses).

### Possibly worth a sweep

`#258` fixed one reader, this is a second. It may be worth grepping for any remaining bare variable inside `$(( ))` that originates from a file, rather than fixing them one report at a time.

### Environment

- plugin `remember` 0.15.0 (`claude-plugins-official` marketplace)
- macOS 26.6, arm64
- `/bin/bash` 3.2.57(1)-release — the shebang's interpreter

jmossie82:
Closes #322

`scripts/save-session.sh` reads two cooldown markers straight into `$(( ))`, where bash evaluates file content as an arithmetic expression. #258 fixed this exact read in `50-git-backup.sh`; these two were missed:

- `tmp/last-save-ts` at the save gate (`:142`)
- `tmp/last-ndc.ts` at the NDC gate (`:565`)

## A correction to my own issue, up front

I filed #322 claiming the script *dies* and that compression then runs on every save. **Both are wrong**, and the issue body and title now carry the correction. Measured on `/bin/bash` 3.2.57 and 5.2.37:

```
enter then-block
t.sh: line 7: 08: value too great for base (error token is "08")
AFTER_FI                     <-- the statements between are all skipped
```

Bash abandons the **entire if/then body** and resumes after `fi`. So:

- **Nothing dies.** `set -e` plays no part — the behaviour is identical with `-e`, without it, and with `-u`.
- **One diagnostic, not two.** `[ "$ELAPSED" -lt … ]` is never reached, so `integer expression expected` is never emitted.
- **Both gates self-heal.** `date +%s > "$COOLDOWN_MARKER"` is below the `fi` (`:158`) and `date +%s > "$NDC_MARKER"` (`:585`) runs on precisely the path a corrupt marker allows.

So the real cost is **one skipped cooldown per corruption event, plus one stderr line** — then it repairs itself. That is a good deal smaller than the issue claimed, and I would rather say so here than have you find it.

**This also corrects something I wrote about your file.** I said the `08` case in `50-git-backup.sh` was #258's death mode. It is not: `$ELAPSED` sits inside the abandoned body, so `set -u` never sees an unbound variable there either. Against the real hook with a corrupt marker: `rc=0`, backup proceeds, commit lands. The consequence is the same in both files.

## Why this is still worth fixing

Narrow, and I will not overstate it:

1. Unvalidated file content should not reach an arithmetic evaluator. [BashPitfalls #7](https://mywiki.wooledge.org/BashPitfalls) covers this read and recommends exactly the `case $foo in ("" | *[!0123456789]*)` form you already chose for #258 — this change just applies it to the two places that were missed.
2. It removes an unexplained diagnostic from `hook-errors.log`, which users can see since #277.
3. ShellCheck does not detect this class at all — verified at 0.11.0 including `-o all`, zero findings. The standing request is koalaman/shellcheck#2679, open and unassigned since 2023-02-01. So a guard in the source is the only thing that catches it.

## `case` and `10#`, not one or the other

`08` and `09` are all digits, so they clear a digits-only guard and are then read as **octal** — `value too great for base`, the identical abandonment, from a marker that looks clean.

```bash
case "$LAST_MOD" in ''|*[!0-9]*) LAST_MOD=0 ;; esac
ELAPSED=$(( $(date +%s) - 10#$LAST_MOD ))
```

`10#` goes after the guard and never instead of it: `10#` on an empty string is itself an error on bash 5 (silently `0` on 3.2).

If the objection is that `10#` is unreachable because no writer emits a leading zero — that is true, and it is equally true of `;`, which is the byte that started this. Neither shape is writer-emitted; both arrive only through a corrupt marker, which is the premise #258 already accepted. I would rather the guard be uniform than draw a line between two kinds of corruption on the grounds that only one of them has been seen.

## A marker that worked and now reads as 0

Disclosing a real behaviour change: `"  1785512249  "` is **not** corrupt today. Arithmetic skips surrounding whitespace, so it currently computes correctly and the throttle works. After this change the `case` rejects it and it falls back to `0`.

Deliberate, and the same call your #258 guard already makes for that file. It is kept as a test parameter so the change is documented rather than discovered.

## Tests

`tests/test_save_session_marker_arithmetic_322.py`, red before / green after.

RED — pristine scripts, this test file added:

```
9 failed, 3 passed
```

GREEN — with the patch: `12 passed`. Full suite: `1517 passed, 43 skipped`, coverage `94.27%` against the 80% floor.

Two deliberate choices in that file:

- **It asserts on the diagnostic, not on whether a save happened.** Falling back to `0` makes a corrupt marker read as "very old", so corrupt and clean-but-ancient reach the same decision — the stderr is the only thing that separates broken from fixed.
- **`garbage-text` is deliberately not a parameter.** It parses as `garbage - text`, two unset identifiers that are both `0` without `set -u`, so it emits nothing and the fix does not change it. A parameter that cannot fail either way is decoration, not coverage. `"  1785512249  "` *is* kept, as the control documenting the change above.

`"integer expression expected"` is likewise absent from the asserted phrases, because the abandoned body means it can never be produced. `"invalid integer constant"` is there instead — that is what bash 5 says if a later change keeps `10#` but drops the `case` in front of it.

## Found in passing — filed as #324

Not in this PR; filed separately as #324 so it stands on its own evidence.

`test_a_corrupt_cooldown_marker_does_not_kill_the_hook` writes `<store>/.last-git-backup-ts`, but the hook resolves the marker through `_gb_common_dir` and reads `<store>/.git/remember/last-git-backup-ts`. I confirmed where the file actually lands by running a backup and globbing for it. **The test has been feeding the script a decoy, so every corrupt value passes trivially and #258's regression is currently unguarded.** The cleanup helper at the top of that file globs both names, which is the trace the path moved.

Pointing it at the real path is not quite enough on its own: with `08` the hook prints the octal error but still returns `rc=0` and commits on bash 3.2.57, so an outcome-only assertion stays green on the macOS leg. Asserting on `proc.stderr` makes it fail on any version.

Patch for that one is ready whenever you want it — see #324.

## Platform

Everything above measured on `/bin/bash` 3.2.57 (macOS, the shebang's interpreter) and cross-checked on 5.2.37. `pytestmark = skipif(sys.platform == "win32", …#79)` matches the neighbouring shell tests.


fdaviddpt:
Merged, and shipping in v0.17.0.

Thank you — genuinely. You filed the issue, you traced the mechanism, and then you sent the patch, which is not the usual order of events around here. A few things I want to name because they were better than they had to be:

- **`case` *and* `10#`, with the reasoning for why it is both and in that order.** The `08`/`09` octal case is the one a digits-only guard quietly lets through, and most patches for this class stop at the digits check.
- **You corrected your own issue in the patch.** The filed text claimed `set -e` and `set -u` consequences; your comment says plainly that both are wrong and gives the measured behaviour instead. That is rarer than the fix.
- **The test asserts the diagnostic rather than whether a save happened**, and the comment explains why — falling back to `0` makes a corrupt marker read as "very old", so corrupt and clean-but-ancient reach the same decision. That reasoning is better than most of what is already in this suite.
- The note about why `garbage-text` is *not* a valid control stayed in, and it should have.

Your #324 is next — you are right that the regression test writes to a path the hook does not read, so that guard is currently asserting nothing.

Two follow-ups came out of auditing your change, both pre-existing and neither yours: #326 (the guards validate syntax but not range, so a future-dated marker sticks the throttle on) and #327 (`10#` landed here only; `50-git-backup.sh` still takes the octal path). Both are credited to your patch for surfacing the class.

Unrelated, but you clearly enjoy this kind of thing: we also maintain [claude-supertool](https://github.com/Digital-Process-Tools/claude-supertool), token-efficient batched ops for Claude Code sessions. Same house style — a checker that cannot answer has to say so rather than return a zero. You would probably find things in it. Issues and PRs very welcome.

Issue #298, opened 2026-08-03
The store's git index and its directory can differ in case; on a case-sensitive restore that may split the store


fdaviddpt:
## Summary

#263/#267/#268 made the drive-letter fold consistent, so a Windows install now resolves its store to `c--Users-…`. Nothing migrates a store that an earlier version created at `C--Users-…`. The old directory keeps the memory written before the upgrade, and no code path reads it again.

`scripts/bootstrap-dirs.sh` has a one-shot migration, but it is legacy `${PROJECT_DIR}/.remember/` → external `REMEMBER_DIR`. There is no migration for *same-layout, differently-cased* store roots. `grep -rn "migrat" scripts/*.sh` finds only the legacy path.

## Evidence this exists in the wild

Reported incidentally by @jqit-ricky on [#297](https://github.com/Digital-Process-Tools/claude-remember/issues/297), while describing their store-root scan:

> We have a store where the directory is spelled `C--Users-ricky-dev-jqit-project-b` while the record inside it says `c--Users-ricky-dev-jqit-project-b`, and matching on `project_dir` returns the spelling the plugin will actually use.

That is one real install holding two spellings, and it is the reporter from #263 — the person whose original bug was twelve days of `nothing to commit … skip`.

## Why this is worth its own issue

The class is **fails to preserve**, not **misreports**: nothing is destroyed, the data is still on the machine, and a rename recovers it. But nothing tells the user, and the failure is shaped so it cannot be noticed from inside:

- NTFS is case-insensitive, so `REMEMBER_DIR` resolves and today's memory saves normally.
- The new store starts empty and fills up. It looks like a working install, because it is one.
- Only the history is gone from view — which is the one thing whose absence looks identical to "we had not written much yet".

Same shape as the defect class this tracker keeps hitting: an absence produced by the tool, read as an absence in the world.

## Unverified, and stated as such

I have not reproduced this. It is a macOS/Linux dev box, and by construction there is no drive letter to get wrong here — that is the same reason #263 needed a real Windows machine. What I did verify: no migration exists for this case, and no open or closed issue covers it (searched `migrate store slug casing`).

So the first question is a measurement, not a design: **how many installs actually hold two spellings?** A store root listing on an affected machine answers it; the reporter of #297 can likely answer for theirs.

## Directions, roughly in order of how little they disturb

[...]

Issue #327, opened 2026-08-08
10# landed in save-session.sh only — 50-git-backup.sh still takes the octal path, where set -u makes it fatal, and a new test's docstring claims otherwise


fdaviddpt:
Found by the release-gate audit of `v0.16.0..main` while cutting v0.17.0.

## The code half

`10#` appears at `scripts/save-session.sh:172`, `scripts/save-session.sh:604` and `scripts/lib-lock.sh:412` — nowhere else in the tree.

`hooks.d/after_save/50-git-backup.sh:172` keeps the bare digits-guard with no `10#`. That file runs `set -u`, which makes the octal path **fatal** rather than merely skipping:

```
$ /bin/bash -c 'set -u; LAST_MOD=08
  case "$LAST_MOD" in ""|*[!0-9]*) LAST_MOD=0 ;; esac
  ELAPSED=$(( $(date +%s) - LAST_MOD )); echo reached; [ "$ELAPSED" -lt 900 ]; echo stamp'
/bin/bash: line 4: 08: value too great for base (error token is "08")
/bin/bash: line 5: ELAPSED: unbound variable
exit=127
```

The hook dies at `:173`, so `_gb_stamp_cooldown` at `:583` is never reached, the marker keeps its `08`, and **the git backup is stopped permanently.** `08`/`09` are all digits, so they clear the guard; they are only reachable through a corrupt marker, which is that guard's whole premise.

`hooks.d/after_save/50-git-restore.sh:225` and `:304` read the same way.

Class: `fails-to-preserve`.

## The docstring half, which is the worse one

`tests/test_save_session_marker_arithmetic_322.py:28-30` states that *"the same `10#` is added to #258's existing guard in `50-git-backup.sh`"*.

It was not. The patch does not touch that file. A reader takes that sentence as coverage of the site that is actually still broken — and it is in a test file, which is the last place anyone re-reads for accuracy.

Class: `misreports`.

## Related, and it is why a blind fix here would be untestable

`test_a_corrupt_cooldown_marker_does_not_kill_the_hook` writes to `<store>/.last-git-backup-ts`, while the hook reads `$GB_STATE_DIR/last-git-backup-ts` where `GB_STATE_DIR="$BACKUP_COMMON_DIR/remember"`. Those paths coincide only on the `REPO_ROOT` fallback, which that fixture does not take.

[...]

Issue #326, opened 2026-08-08
The cooldown guards validate syntax but not range: a digits-only marker ahead of now sticks the throttle ON permanently, with no diagnostic


fdaviddpt:
Found by a release-gate security audit of `v0.16.0..main` while cutting v0.17.0. Pre-existing rather than introduced by #322/#325 — but it is the missing half of the guard that release exists to add, and the release note asserts the opposite of it.

## Mechanism

`case "$LAST_MOD" in ''|*[!0-9]*) LAST_MOD=0 ;; esac` accepts **any** all-digit string. It rejects the syntax error; it does not bound the value.

A marker holding a future epoch — clock set back after an NTP correction, a VM snapshot restore, a container clock jump — or an overflowing integer produces a negative `ELAPSED`. `[ "$ELAPSED" -lt "$SAVE_COOLDOWN" ]` is then **true**, and the `exit 0` fires.

The `exit 0` sits **above** the line that rewrites the marker. So the self-heal that the whole fix rests on is unreachable on exactly the path that needs it:

```
$ /bin/bash -c 'for M in 99999999999999999999 9223372036854775807; do
    LAST_MOD=$M; case "$LAST_MOD" in ""|*[!0-9]*) LAST_MOD=0;; esac
    E=$(( $(date +%s) - 10#$LAST_MOD )); printf "marker=%-22s ELAPSED=%s " "$M" "$E"
    if [ "$E" -lt 120 ]; then echo "-> COOLDOWN ENGAGED (exit 0, marker NOT rewritten)"; else echo "-> proceeds"; fi; done'
marker=99999999999999999999   ELAPSED=-7766279629666055955 -> COOLDOWN ENGAGED (exit 0, marker NOT rewritten)
marker=9223372036854775807    ELAPSED=-9223372035068589843 -> COOLDOWN ENGAGED (exit 0, marker NOT rewritten)
```

Every subsequent run skips, permanently, until the wall clock catches up or someone runs `--force` or deletes the file by hand. **No stderr, nothing in `hook-errors.log`.**

## Four sites, and on each the rewrite is unreachable on the stuck path

| Site | Stuck state | Rewrite that never runs |
| --- | --- | --- |
| `scripts/save-session.sh:169-177` | no session is ever saved again | `:187` |
| `scripts/save-session.sh:600-604` | NDC compression never runs again; `now.md` grows unbounded | `:622`, inside `if [ "$RUN_NDC" = true ]` |
[...]

PR #329, opened 2026-08-08
fix(#322): the third reader of the same marker, and a freeze guard that outlived its claim

fdaviddpt:
Builds on @jmossie82's #325 (merged). Their patch is unchanged — this is the sweep it opened up.

Part of #322

## The third and fourth sites

`10#` landed in `save-session.sh`. `scripts/post-tool-hook.sh` reads the same `tmp/last-save-ts` at `:335` and `tmp/no-transcript-notice` at `:192`, both `case`-guarded since #230, neither carrying `10#`. So `08`/`09` clear the guard and are read as octal, the body is abandoned, and:

- `:335` — `IN_COOLDOWN` stays false, the hook forks a save inside the window
- `:192` — worse: the abandoned body swallows its own `exit 0`, so the line below the `fi` deletes the notice marker on the assumption a transcript was found

Both tests assert **behaviour** (`save-session.pid` absent / the notice marker surviving), not just the diagnostic. That mattered: the first draft asserted `proc.stderr` alone and **passed against the broken code**, because bash's line goes to `hook-errors.log` rather than stderr on this path.

## `SAVE_COOLDOWN` was left alone, deliberately

I briefed this as having the same gap. It does not. `LAST_TS` is inside `$(( ))`; `SAVE_COOLDOWN` is the right-hand operand of `[ … -lt … ]`, the `test` builtin, which parses base 10 and does not evaluate:

```
$ [ 9 -lt 010 ] && echo decimal || echo octal      -> decimal
$ X='a[$(touch p3)]'; [ 5 -lt "$X" ]              -> integer expression expected, rc 2, nothing run
```

`10#` there would be inert and worse than inert — it would advertise a hazard at a site that has none, and its absence elsewhere would then read as a finding. The measurement is in the code comment so it does not get "fixed" later.

## The judgment call that needs a second opinion

`test_the_per_tool_call_path_is_not_touched` (#298/#299) asserted `post-tool-hook.sh` is **byte-identical to origin/main**. That blocked a correct one-token fix and would block every future edit to the file. It is also self-cancelling: once the edit it objects to lands on main, it compares against the new main and passes. So what it actually asserts is *"nobody has touched this file since main"* — a different claim from the one its name makes, *"the per-tool-call path is still cheap"*. A check that reads as evidence about the world when it is evidence about the diff.

Replaced the `post-tool-hook.sh` arm with the three properties #298/#299 care about — the divergence library is not sourced, `case_divergence` is not called, no `git` is spawned — each with its own message. Stronger than byte-equality after this merges, and cannot go vacuous. The `lib-slug.sh` / `lib-memory-dir.sh` arms keep their byte compare untouched.

**If you would rather that were its own PR against #298, say so** — but then the third site cannot land at all, because the guard forbids it.

## Verification

Their tests against main's `save-session.sh`: `9 failed, 3 passed` (the 3 are the two whitespace controls and the valid-marker case, by design). Their tests with their patch: `12 passed`. My two, before my `10#`: red with `value too great for base` at `:325` and `:186`. Full suite `1519 passed, 43 skipped`.

macOS + bash 3.2.57 only. **bash 5 unverified** — no bash 5 on this machine. CI is the authority.


fdaviddpt:
### Code review

No blocking issues. One non-blocking finding, filed as #330 rather than held against this PR.

The byte-identity pin on the per-tool-call path is gone, and the three substring checks that replace it can be walked around — `GIT=git; "$GIT" -C ...`, an inlined divergence check without the literal `case_divergence`, or any added spawn or file read that is not a `git` call. #299 cited that byte-identity as its evidence for the zero-cost claim. Some loosening was unavoidable, since the old assertion blocked this PR's own one-character fix, so the right answer is a cost-counting pin rather than keeping the byte compare. Scored 60. Details in #330.

Checked and clean: the two `10#` sites are both already `case`-guarded to digits, so `10#` never meets an empty string; `SAVE_COOLDOWN` is correctly left alone as `test`'s right operand; the `50-git-backup.sh` / `50-git-restore.sh` deferral matches #327's ordering; bash 3.2.57 compatibility holds across the changed lines; no comment or docstring left contradicting the code.

Adjacent, already open: #326 names `post-tool-hook.sh:316-326` — the same line this PR touches — for a range check rather than a syntax check. A future-dated marker sticks the throttle on with no self-heal. Out of scope here, still open there.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

Issue #303, opened 2026-08-04
A spawn-count test can measure the cold path by accident: -nt is second-granular, so writing a config races the cache


fdaviddpt:
Bash's `-nt` compares modification times at **whole-second** granularity. The env cache is invalidated by comparing the cache file against the config files with `-nt`, so a config written in the *same second* as the cache reads as "not newer" and the cache is rejected.

That direction is safe — one extra re-resolution, never a stale answer — and this issue does not ask to change it.

## The hazard is in the tests, not the product

A test that writes a config and then counts process spawns on a "warm" run is measuring one of two different things depending on which side of a second boundary the two writes landed on:

- different seconds → cache rejected → the run is **cold** → spawn count is high
- same second → cache honoured → the run is **warm** → spawn count is low

Both outcomes are correct product behaviour. Only one of them is the thing the test claims to measure. A warm-path spawn budget asserted this way passes or fails on timing, and when it passes there is nothing in the output distinguishing "the warm path is cheap" from "the run happened to be warm".

## Why nobody has hit it yet

The existing suite dodges it by accident rather than by design:

- `test_a_warmed_prompt_hook_spawns_almost_nothing` writes **no config at all**, so there is no second write to race.
- the timezone test backdates its config by +5s, for its own unrelated reasons.

Neither of those is a guard, and neither generalises. The #301 spawn test backdates explicitly with the reason in a comment — which fixes that one test and leaves the trap in place for the next person who writes one.

## What would settle it

A shared fixture for "write a config and make it unambiguously older than the cache" (or newer, when that is the point), so a spawn-count test cannot accidentally measure the cold path. The knowledge currently lives in one comment in one test file, which is where it will stay until someone rediscovers it.

## Class

*Misreports* — nothing is lost and nothing is broken; a measurement can simply report a number that answers a different question than the one asked, with no way to tell from the output which question it answered.

Found by the agent implementing #301, reported unprompted as an out-of-scope observation.

Issue #301, opened 2026-08-03
user-prompt-hook's per-turn timestamp defeats prompt prefix caching — add a toggle or drop the clock


jackneil:
## Problem

`scripts/user-prompt-hook.sh` prints `[$(date '+%H:%M %Z') — $(whoami)]` (or the `— ${CTX_PCT}%` variant) to stdout on every UserPromptSubmit — currently lines ~156-158 at HEAD:

```sh
_REMEMBER_NOW=$(_remember_date '+%H:%M %Z')
TIMESTAMP="[$_REMEMBER_NOW — $_REMEMBER_WHO — ${CTX_PCT}%]"
```

Claude Code injects hook stdout into the model prompt as `UserPromptSubmit hook success: [13:05 EDT — jack.neil]`. Prompt prefix caches — Anthropic's server-side prompt caching and self-hosted gateways alike — match byte-for-byte from the front, so a wall-clock stamp inside the prompt forces re-prefill of everything positioned after it whenever the minute ticks over between turns or sessions.

Measured impact (2026-08-03, self-hosted DeepSeek-V4 gateway, ~62K-token Claude Code openers): this line was one of two volatile byte sources capping cross-session prefix reuse; on large local models the re-prefill it triggers costs real minutes per day. Because there's no config toggle, gateway operators end up regex-rewriting the line in flight, which is fragile for everyone.

## Request

Any of these would fix it cleanly:

1. A config toggle to suppress the stamp entirely, or
2. Emit it without the clock — `[$(whoami)]` and the context-percentage warning are the useful parts and are byte-stable between turns (the `CTX_PCT` warning could stay threshold-gated as today, so it only appears when it changes behavior anyway), or
3. Emit the clock only when it materially changes (e.g. hour granularity).

Happy to PR whichever shape you prefer.

Related: #227 touches the same hook for spawn-cost reasons; if that lands a rewrite, this would be a natural rider.

Issue #232, opened 2026-07-30
post-tool-hook: the last 17 spawns are the merged config, and caching it is a credential-lifetime decision


fdaviddpt:
Split out of #230, which took `post-tool-hook.sh` from **30 external spawns per tool call to 17** (16 in a real git checkout) without adding any caching. This is the remainder #230 deliberately did not take.

## What is left, measured

Warm run, macOS bash 3.2.57, PATH-shim counter (`tests/test_post_tool_hook_spawns.py` carries the harness):

| count | command | source |
|---|---|---|
| 1 | `jq -s` (three-layer deep merge) | `lib-memory-dir.sh` |
| 3 | `jq -r` (`.timezone`, `.model`, `.reject_pattern`) | `log.sh`, at source time |
| 2 | `jq -r` (`.cooldowns.save_seconds`, `.thresholds.delta_lines_trigger`) | `post-tool-hook.sh` |
| 1 | `rm -f` (merged config, EXIT trap) | `lib-memory-dir.sh` |
| 1 | `sed` (`trap -p` parsing, to chain the EXIT trap) | `lib-memory-dir.sh` |
| 1 | `sed` (log message sanitisation) | `log.sh` |
| 2 | `date` (`+%Y-%m-%d`, `+%H:%M:%S`) | `log.sh` — **bash 3.2 only**, `lib-clock.sh`'s builtin covers bash ≥ 4.2 |
| 1 | `date +%s` | routed to `date` on every bash: `%s` is a strftime extension `lib-clock.sh` does not trust the builtin with |
| 1 | `python3 -V` (validate the interpreter) | `detect-tools.sh` |
| 1 | `git rev-parse` | `lib-memory-dir.sh` — **only when `PROJECT_DIR` has no `.git` directory**; 0 in an ordinary checkout |
| 1 | `ls -t` + 1 `head -1` | `post-tool-hook.sh`, newest-transcript pick |
| 1 | `wc -l` | line count of the transcript |
| 1 | `mv -f` | atomic `capture-alive` write |

**Seven of the seventeen are the merged config**: the `jq -s` merge, five `jq -r` reads against it, and the `rm` that cleans it up. That is the single biggest remaining block and the reason this is filed rather than fixed.

## Why #230 did not cache it

[...]

Issue #206, opened 2026-07-27
False "previous session was not captured" warning fires on every /clear


fdaviddpt:
## Symptom

Right after `/clear`, the next user prompt is annotated with:

> remember: your previous session was not captured. If you just installed or enabled the plugin, that is expected — capture starts now. Otherwise its hooks were not registered for that session; run /remember:doctor.

Capture was healthy. Observed 2026-07-27 23:08 in the DVSI project; at that moment `.remember/now.md` had been written 23:07:48 and `today-2026-07-27.md` 22:50:53. Nothing was missed.

## Mechanism

`scripts/session-start-hook.sh:139-148`:

```sh
PREV_JSONL=$(ls -t "$(claude_projects_dir)/${PREV_SLUG}"/*.jsonl 2>/dev/null | tail -n +2 | head -1)
PREV_ID=$(basename "$PREV_JSONL" .jsonl)
SEEN_ID=$(cat "$CAPTURE_ALIVE" 2>/dev/null)
if [ -n "$PREV_ID" ] && [ "$SEEN_ID" != "$PREV_ID" ] && grep -q '"tool_use"' "$PREV_JSONL"; then
    # warn
fi
```

The check assumes SessionStart means "a new session id started, so `capture-alive` still holds the previous one".

**`/clear` does not mint a new session id.** It fires SessionStart with `source=clear` while the transcript, the session id, and the `.jsonl` file all stay the same. Confirmed in the incident: the newest `.jsonl` (`b0a76ca7…`, 513K — far too large to be a just-created session) is the *current* session, and `tmp/capture-alive` already contained `b0a76ca7…` because that session had made tool calls **before** the clear. `PREV_ID` resolved to the genuinely older `c1ab365b…`.

So on a `/clear`, `SEEN_ID != PREV_ID` is structurally guaranteed:

| | before /clear | at SessionStart(source=clear) |
|---|---|---|
| newest jsonl | current session | current session (unchanged) |
| `capture-alive` | current session id | current session id |
| computed `PREV_ID` | — | second-newest = an older session |

The comparison can never hold, whatever the capture state. Same reasoning applies to `compact` and `fork` — every same-session SessionStart source.

## Root enabler

[...]

PR #352, opened 2026-08-15
perf(post-tool): give PostToolUse the #227 env-cache fast path

### LINKED ISSUE #350: post-tool-hook.sh: apply the #227 env-cache fast path to PostToolUse (~0.75-1.0 s per tool call on Windows)
**Versions checked:** 0.19.0, 0.20.0, and current `main` HEAD (post-tool-hook.sh structure identical in all three)
**Environment:** Windows 11 Pro, Git Bash (bash 5.x), Claude Code 2.1.232, plugin installed from claude-plugins-official

## Summary

`post-tool-hook.sh` registers with no matcher, so it runs on **every tool call** — and each invocation pays for the full bootstrap chain plus at least one Python interpreter start. On Windows that is 0.75–1.0 s of silent, blocking wait per tool call. `user-prompt-hook.sh` already solved exactly this problem with the #227 env-cache fast path; `post-tool-hook.sh` never got it.

## Measurements

Invoked exactly as the hook system does (`bash scripts/post-tool-hook.sh` with a realistic PostToolUse JSON payload on stdin, real session id, `last-save.json` present). 10 runs each:

| Scenario | min | median | max |
|---|---|---|---|
| post-tool-hook, `python3` = Microsoft Store WindowsApps alias | 947 ms | **1,010 ms** | 1,100 ms |
| post-tool-hook, `python3` = real CPython 3.14.7 (0.19.0) | 730 ms | **744 ms** | 955 ms |
| post-tool-hook, same, 0.20.0 (5 runs) | 771 ms | **793 ms** | 806 ms |
| user-prompt-hook.sh (has the #227 fast path), warm | 66 ms | **~90 ms** | 145 ms |
| Baseline `bash -c ':'` spawn | 23 ms | 25 ms | 33 ms |
| `python3 -c pass` (Store alias / real) | — | 125 ms / 36 ms | — |
| `python3 -m pipeline.shell read-position` | — | ~165 ms (Store), ~90 ms (real) | — |
| `jq --version` spawn | — | ~34 ms | — |

The dominant cost is not any single process but the count: unconditional `source resolve-paths.sh → detect-tools.sh → bootstrap-dirs.sh → log.sh` on every call (detect-tools spawns an interpreter just to validate `-V`), jq config reads, `ls -t | head`, `wc`, plus `$PYTHON -m pipeline.shell read-position` once `last-save.json` exists.

At ~1,500 tool calls on a busy agentic day this is **~19–25 minutes/day of blocked agent time**, invisible in Claude Code transcripts because successful hooks with empty output are never persisted.

## You already fixed this once

`user-prompt-hook.sh`'s header (#227) documents the identical Windows story (per-spawn 150–800 ms, p50 8,718 ms before the fix) and its `lib-env-cache.sh` replay path is why it measures ~90 ms warm on the same machine. `post-tool-hook.sh` is the same shape of hook — hotter, even, since tool calls outnumber prompts ~10:1 in agentic sessions — but still takes the full chain every time.

## Proposal

1. **Apply the #227 pattern to post-tool-hook.sh**: when `_remember_env_cache_load` succeeds and `hooks.d/after_post_tool` has no executable listener, skip resolve-paths/detect-tools/log sourcing entirely and run the delta/cooldown logic with replayed `REMEMBER_DIR`/`PYTHON` values.
2. **Keep Python off the hot path**: have `save-session.sh` also write the per-session position to a bash-`read`-able sidecar (e.g. `tmp/position.<session_id>`, plain integer), so `pipeline.shell read-position` only runs when the sidecar is missing — mirroring the `read`-not-`cat` discipline the script already applies to its other markers (#230).

Expected result based on the user-prompt-hook numbers: ~750–1,000 ms → well under 150 ms warm.

Happy to re-run the measurement matrix on a branch.

fdaviddpt:
Closes #350 (part 1 only — the `save-session.sh` position sidecar is deliberately not here; see "What is not in this PR" below).

## The problem

`scripts/post-tool-hook.sh` registers with **no matcher**, so it runs after every single tool call and the agent blocks on it. Every invocation sourced the whole chain unconditionally — `resolve-paths.sh` → `detect-tools.sh` → `bootstrap-dirs.sh` → `log.sh` — which is a `git rev-parse`, a slug, a three-layer config merge, a one-pass flatten, and a `python3 -V` spent only to validate an interpreter.

The reporter measured **750–1000 ms per tool call** on Windows 11 / Git Bash, against ~90 ms for `user-prompt-hook.sh` on the same machine — the hook that already replays. Their measurement is a claim I did not reproduce (no Windows box); I re-measured the shape on macOS and it holds there too.

## The change

The #227 fast path now covers this hook. When `lib-env-cache.sh` can replay a published resolution **and** no *executable* listener is installed under `hooks.d/after_post_tool/`, the chain is skipped.

Observed, macOS / bash 3.2.57, external spawns per warm tool call:

| fixture | before | after |
|---|---|---|
| no save yet | 14 spawns / 336 ms | **6 / 130 ms** |
| a save behind it | 15 spawns / 405 ms | **8 / 248 ms** |

The first tool call of a session — and the first after any config edit — still takes the whole chain, and now publishes it, so the cost is paid once per project per config change instead of per tool call.

## The objection #227 raised, answered rather than dropped

`tests/test_post_tool_hook_spawns.py` says in its own docstring why #227 skipped this hook: it needs `config()`, and therefore the merged config file, "caching which is a materially bigger risk than caching a handful of paths". That is right, and the merged config file is **still never cached** — it can carry a live `haiku.oauth_token`, which is why `lib-memory-dir.sh` creates it `0600` per PID under an `EXIT` trap (#232/#68).

What is cached are the two **scalars** `log.sh` resolves out of it — `cooldowns.save_seconds` and `thresholds.delta_lines_trigger` — in the same `0600` file that has carried `REMEMBER_TZ` since #227, under the same config-mtime invalidation, with the same "required, not defaulted" upgrade rule #301 gave `REMEMBER_PROMPT_STAMP`. A cooldown and a line threshold are neither secret nor expensive to be one prompt stale about.

## Four things that were easy to get wrong

**The gate asks for an executable, not a directory.** `user-prompt-hook.sh` tests `[ ! -d hooks.d/after_user_prompt ]` and gets away with it because the distribution ships no such directory. It **does** ship `hooks.d/after_post_tool/.gitkeep` (checked: `git ls-files hooks.d`), so the same `-d` test here would have refused the fast path for every user who never installed a listener — which is all of them. It asks what `dispatch()` asks, from one shared definition, so the gate and the dispatch cannot come to disagree about whether a listener exists.

**`PYTHON` is not replayed.** `lib-env-cache.sh` validates config mtimes and environment identity; it does not validate binaries, and it was not asked to. `detect-tools.sh` is sourced at the one site that needs an interpreter — reading the saved position, which a store only reaches once a save has landed — so the Microsoft Store alias that passes `command -v` and fails `-V` is still caught by exactly the check that has always caught it.

**`log` is upgraded, not stubbed.** `dispatch` is stubbed, because the gate only let us in when there is nothing to dispatch to. `log` is different: the branches that call it are the ones where this hook is *malfunctioning* — a slug matching no session directory (#144) — and dropping those diagnostics to save a process is #144 arriving inside its own fix. The first call sources `log.sh`; the runs that never take those branches never pay for it.

That upgrade has to install a no-op when `log.sh` cannot be sourced at all, and the obvious guard — `type log` — is **true on macOS regardless**, because `/usr/bin/log` is Apple's unified logging CLI. Found by running it: the hook execed that binary once per diagnostic, printed nineteen lines of `log: Unknown subcommand 'hook'`, and exited **64** from a hook documented "EXIT CODES: 0 Always". It asks `declare -F`.

**The #200 wiring marker survives a store with no `tmp/`.** `bootstrap-dirs.sh` is what creates `$REMEMBER_DIR/tmp`. `tmp/post-tool-ran` is how `/remember:doctor` answers "is `PostToolUse` wired at all", so it now creates the directory on the failing write rather than assuming one. Silently skipping it would tell users a wired hook had never fired — the exact regression #200 fixed.

Also done by hand, because the fast path skips whatever set them: the #204 nested-summarizer guard, `umask 077` (#68), `SYS_TMPDIR`, `REMEMBER_HOOKS_DIR`, and `bootstrap-dirs.sh`'s `exec 2>>` into `hook-errors.log` — without that last one a diagnostic lands in front of the user instead of in the log.

## Tests

`tests/test_post_tool_fast_path_350.py`, 13 cases, written before the implementation and watched failing (9 red, 2 green-for-the-right-reason). Every "the chain did not run" assertion is paired with a positive control in the same fixture:

- the spawn budget, with the cold run asserted to spawn `git` and `jq` (so "no git, no jq" on the warm run is evidence of something) and the warm run asserted to spawn `ls` (so it is evidence of a hook that ran);
- the threshold and the saved position, each parametrised **both ways** — a value that must suppress the fork and one that must produce it, because only asserting the first passes on a hook that stopped forking altogether;
- the #144 diagnostic, with the inverse control that a healthy run says nothing;
- the #200 marker with `tmp/` deleted, the #206 per-session marker, the #204 guard with a control run beside it, `.gitkeep` not disabling the fast path, an executable listener taking the slow path and being dispatched, and clean stdout/stderr.

Two of them have their negative controls recorded rather than assumed: reverting `log()` to a stub and reverting `declare -F` to `type` each make the corresponding test fail, and I ran both.

A helper (`_fast_path_run`) establishes fast-path-ness from the spawn log rather than from `EnvCacheProbe.assert_warm`, because against a cache primed by another hook the probe answers `warm` for the slow path too — it cannot tell "replayed" from "never looked".

**Two adjacent tests were measuring paths they did not name, and now say which.** `tests/test_post_tool_hook_spawns.py` counted a *second* run of the same fixture, which after this change is the warm one — and whether it was warm at all depended on which side of a whole second two writes landed on (#303). It pins `REMEMBER_ENV_CACHE=0` and is now explicitly the cold-path file. `TestFreshProjectBootstrap::test_detect_before_bootstrap` compared the first *mention* of two filenames anywhere in a hook, comments included; it now reads the `source` statements, with the premise (both are in fact sourced) asserted rather than assumed. `test_case_divergence_298` pinned two hot-path libraries to `origin/main` byte for byte and so failed on a comment — the very comment that had become false; it now compares code lines only, verified still to catch a renamed function.

Full suite: **1590 passed, 43 skipped**, coverage 92.95%.

## What is not in this PR

Part 2 of the issue — having `save-session.sh` write a plain-integer `tmp/position.<session_id>` sidecar so `pipeline.shell read-position` stays off the hot path. It is worth about 2 spawns and ~118 ms per warm tool call on my machine, and it introduces a second source of truth for a position `last-save.json` already holds; a sidecar that disagrees with it produces a wrong delta silently. That is a separate argument and deserves its own issue.

## Platform

Everything measured above is **observed** on macOS 24.3.0 / bash 3.2.57. Every bash test in this repo skips on `win32`, so the Windows CI leg does not execute any of this, and no Windows machine was available.

One claim is **reasoned, not observed**, and is written down above `_after_post_tool_listener` in the code: MSYS fakes the execute bit, and if Git Bash ever reported the shipped mode-`0644` `.gitkeep` as executable, the fast path would never fire on Windows. That would be slow, not wrong — the slow path is today's behaviour exactly — and `dispatch()` has shipped the identical `[ -x ]` test for many releases, so it would already be trying to *execute* that file on every Windows tool call, which nobody has reported. One command in a real Git Bash session settles it: `[ -x hooks.d/after_post_tool/.gitkeep ] && echo YES`.

Also reasoned: CRLF in the cache file is stripped before the new digits-only guards, so a corrupted value falls back to the chain rather than reaching `$(( ))`. `SYS_TMPDIR` and `umask 077` are byte-identical replays of the expressions already in `bootstrap-dirs.sh` and `resolve-paths.sh`.


fdaviddpt:
Part 2 of #350 — the `tmp/position.<session_id>` sidecar — is filed as #353 with the measurement it is worth (~2 spawns / ~118 ms per warm tool call, observed on macOS) and the argument against shipping it as-is. This PR closes #350; #353 is where that half continues.

Issue #432, opened 2026-08-29
The hook-leak security tests skip on Windows although a working Git Bash runner already exists in the suite


fdaviddpt:
Found by lane 424's audit spawn. Not fixed there: the gap is inherited from the #417 precedent that lane deliberately mirrored, and closing it means reworking a shared subprocess-test pattern across at least two files.

## Symptom

`tests/test_transcript_path_leak_424.py` -- the regression test for a release-blocking security fix -- never executes on the `windows-latest` CI leg. Nor does `tests/test_hook_cwd_leak_417.py`, its precedent. Both carry a blanket `pytestmark = pytest.mark.skipif(sys.platform == "win32")`.

The matrix runs Windows. The tests are collected and skipped, so the leg is green and the coverage is zero -- a check that never ran rendering exactly like a check that found nothing.

## Location

`tests/test_transcript_path_leak_424.py` and `tests/test_hook_cwd_leak_417.py`, module-level skip.

`tests/test_hooks_json.py` already ships `_find_git_bash()`, which locates a real bash on Windows and runs bash-subprocess hook tests there for real.

## Mechanism

The skip was copied from a precedent rather than derived from a limitation. `_find_git_bash()` existing in the same suite is the evidence that the limitation is not real -- these tests spawn a hook through bash exactly as `test_hooks_json.py` does.

What the skip costs specifically: the `unset` in `post-tool-hook.sh` and `user-prompt-hook.sh` is a bash-level guarantee, and Git Bash on Windows is a different bash with different environment-inheritance behaviour from the macOS and Linux legs. That is precisely the platform band this repository's CLAUDE.md warns about -- a green run on the platform the code was written on says almost nothing about the one it was not.

## What would settle it

Move both files onto `_find_git_bash()` (or a shared helper extracted from it) and narrow the skip to the case where no bash is found, so that a missing interpreter reports as skipped-for-cause rather than as a platform-wide exemption. If the two tests genuinely cannot run under Git Bash for a reason nobody has written down yet, that reason belongs in the skip's own message.

PR #680, opened 2026-09-12
Fix NUL-interleaved UTF-16LE bash probe output by using resolve_bash() and defensive decode

### LINKED ISSUE #672: windows-latest: a bash subprocess probe's output arrives NUL-interleaved (UTF-16LE), regardless of invocation shape
Part of #660/#669 (PR #671). While landing #669's Windows Git-Bash benchmark, a regression test that spawns a tiny `bash` subprocess to probe `session-start-hook.sh`'s own `_HOOK_DIR="${BASH_SOURCE[0]%/*}"` derivation (`tests/test_session_start_windows_benchmark_669.py:577`) consistently fails on every `windows-latest` CI leg (3.9/3.10/3.11/3.12) with `returncode == 1` and empty `stderr`, but the `subprocess.run(...)` call's own repr shows corrupted, NUL-byte-interleaved text where real content should be.

Verbatim, from a round-7 job log (windows-latest, Python 3.9/3.10, job ids 103483869911 / 103483869958, run 34668023792):

```
E        +  where 1 = CompletedProcess(args=['bash', 'C:/Users/runneradmin/AppData/Local/Temp/pytest-of-runneradmin/pytest-0/test_hook_dir_d...\x00s\x00t\x00r\x00o\x00>\x00'\x00 \x00t\x00o\x00 \x00i\x00n\x00s\x00t\x00a\x00l\x00l\x00.\x00\n\x00\n\x00', stderr='').returncode
```

Reading every other byte past the NULs spells roughly `...stro>' to install.` -- the classic byte pattern of ASCII text that was actually UTF-16LE-encoded and then decoded one byte at a time as if it were UTF-8/Latin-1 (each `\x00` byte is independently a valid UTF-8 codepoint, U+0000, so `bytes.decode("utf-8")` does not raise here -- it silently produces a string with embedded NUL characters).

## Why this is environment-level, not this test's own argv construction

This exact corrupted tail text recurred **verbatim**, unchanged, across two structurally different ways of invoking the same bash probe:

1. Round 5: `subprocess.run(["bash", "-c", f'FAKE_SOURCE="$1"; {snippet}; ...', "--", fake_source], ...)` -- an inline script passed as a quoted argv element, with a second argv element carrying the fake path.
2. Round 7 (the log quoted above): `subprocess.run(["bash", probe_script.as_posix()], env=env, ...)` -- a real script file passed as a plain path, no inline quoting, no embedded-quote argv element, and the fake value passed via an environment variable instead of argv.

Round 6 diagnosed (1) as a `list2cmdline`/MSYS argv-requoting mismatch on a string carrying embedded double quotes (citing prior art in `tests/test_fast_path_subshells_511.py`) and 'fixed' it by switching to shape (2). That fix did eliminate shape (1)'s failure mode locally and appeared to work -- CI still failed identically, with the same corrupted tail, on shape (2), which has no embedded quotes to mis-requote at all. That rules out the argv-quoting theory as the (sole) cause.

## Likely suspects (not yet confirmed -- no Windows host available to this investigation)

- A PATH entry resolving `bash`, `python`, or some tool in the probe's own shebang/invocation chain to a **Windows Store Python app-execution-alias stub** rather than a real interpreter -- those stubs are known to write unexpected UTF-16 framing to a pipe when their own stdout is redirected/piped rather than attached to a real console.
- A **Chocolatey shim** (`.exe` shim wrapping a real tool) on `windows-latest`'s default PATH that proxies stdout/stderr and re-encodes it, which would explain why the corruption appears regardless of how the *test* constructs its own command -- something further down the invocation chain (inside what `bash` itself execs, or a wrapper bash resolves to) is the actual source.
- GitHub's `windows-latest` runner image console/codepage configuration interacting badly with `subprocess.run(..., text=True, capture_output=True)`'s own universal-newlines/locale decoding when the child process (or something it shells out to) writes UTF-16.

## What to do

- Reproduce on an actual Windows box (not available to the lane that found this) -- capture `result.stdout`/`result.stderr` as raw bytes (`capture_output=True` without `text=True`) to see the true byte sequence before any decoding, and `where bash`/`where python` to see what's actually resolving.
- Once the real source is identified, decide whether it is fixable in this repo's own test harness (decode defensively, strip/handle the corruption) or whether it is purely an artifact of this CI image that should just be worked around.
- `tests/test_session_start_windows_benchmark_669.py:577`'s own regression test (`test_hook_dir_derivation_needs_a_forward_slash_path`'s backslash-path case) is marked `skipif sys.platform == "win32"` for now, referencing this issue, so the rest of #669's actual benchmark (which passes cleanly on every Windows leg since round 5) is not held up by a harness-only test's own platform quirk.

[AI-generated]

louisss1016:
Closes #672

## Problem
`test_hook_dir_derivation_needs_a_forward_slash_path` (tests/test_session_start_windows_benchmark_669.py:577) failed on every `windows-latest` CI leg: the probe invoked the literal `"bash"`, which on Windows resolves to the WSL launcher (or a WindowsApps alias stub) instead of Git Bash. Those launchers write UTF-16LE-framed text to a redirected pipe (each ASCII byte followed by a NUL), which locale decoding turns into NUL-interleaved garbage — the `...stro>' to install.` byte pattern in the issue's job log is the WSL launcher's missing-distribution message.

## Root cause (reproduced on a Windows host)
`where bash` → `C:\WINDOWS\system32\bash.EXE` (WSL launcher). Its redirected stderr is UTF-16LE, and Windows paths (`C:/...`) are unresolvable inside WSL (rc=127, "No such file or directory"). With `text=True` on a UTF-8 locale the decode raises outright; on CI's locale it silently produced NUL-interleaved text. The test's own `_probe` bypassed `resolve_bash()` — the repo's existing Git-Bash locator (#432) that every other Windows test in this repo uses.

## Fix
- The probe now invokes `resolve_bash()`'s Git Bash instead of the literal `"bash"`, so the WSL launcher is never reached.
- Output is captured as raw bytes and decoded via the new `decode_bash_output()` helper, which detects a UTF-16LE frame by NUL density and falls back to UTF-8-with-replacement.
- The `skipif(sys.platform == "win32")` guard referencing #672 is removed — the harness defect is fixed, not worked around.

## Changes
- `tests/_bash_runner.py`: add `decode_bash_output(raw: bytes) -> str`
- `tests/test_session_start_windows_benchmark_669.py`: probe + real-probe use `BASH` with defensive decode; #672 skipif removed
- `tests/test_decode_bash_output_672.py`: unit tests (UTF-16LE positive control, UTF-8 passthrough, empty, undecodable fallback)
- `changelog.d/672.fixed.md`: changelog fragment

## Testing
Environment: Windows host (no Git Bash installed; `bash` resolves to the WSL launcher), Python 3.12.6, pytest 9.1.1.

- Baseline: `python -m pytest -o addopts="" -q` → **13 failed, 851 passed, 1578 skipped** (13 pre-existing WSL-bash environment failures in test_shell_parse_gate.py / test_safe_eval_seam.py, unrelated to #672)
- After fix: same command → **13 failed, 855 passed, 1578 skipped** — identical failure set, **zero new failures**, +4 new passing unit tests
- Unit: `python -m pytest tests/test_decode_bash_output_672.py -o addopts="" -q` → **4 passed**
- `python -m py_compile` on all three touched modules → clean

Note: `windows-latest` CI ships Git for Windows, so `resolve_bash()` returns a real Git Bash there and this test now runs for real; hosts without Git Bash hit the pre-existing module-level "no usable bash" skip (#432). The UTF-16LE decode path is covered deterministically by the unit tests on every platform.

### DIFF
--- a/changelog.d/672.fixed.md
+++ b/changelog.d/672.fixed.md
@@ -1,0 +1,1 @@
+- Fixed the Windows Git-Bash benchmark's hook-dir probe: it invoked the literal `bash` (which resolves to the WSL launcher and emits UTF-16LE/NUL-interleaved output, exit 127 on a missing distribution) instead of `resolve_bash()`'s Git Bash, and its output decoding is now UTF-16LE-aware (#672).
--- a/tests/_bash_runner.py
+++ b/tests/_bash_runner.py
@@ -62,3 +62,28 @@ def resolve_bash() -> str | None:
     if sys.platform != "win32":
         return shutil.which("bash")
     return find_git_bash()
+
+
+def decode_bash_output(raw: bytes) -> str:
+    """Decode a bash subprocess byte stream defensively (#672).
+
+    On Windows, invoking the literal ``"bash"`` can resolve to the WSL
+    launcher or a WindowsApps alias stub instead of Git Bash. Those
+    launchers write UTF-16LE-framed text to a redirected pipe -- each
+    ASCII byte followed by a NUL -- which a locale/UTF-8 decode turns into
+    NUL-interleaved garbage, or raises outright on a UTF-8 locale. Detect
+    a UTF-16LE frame by its NUL density and decode it as UTF-16LE;
+    otherwise fall back to UTF-8 with replacement, the same policy
+    ``subprocess`` itself uses for undecodable bytes when ``text=True``.
+    """
+    if not raw:
+        return ""
+    if raw.count(b"\x00") >= len(raw) // 4:
+        try:
+            decoded = raw.decode("utf-16-le")
+        except UnicodeDecodeError:
+            pass
+        else:
+            if "\x00" not in decoded:
+                return decoded
+    return raw.decode("utf-8", errors="replace")
--- a/tests/test_decode_bash_output_672.py
+++ b/tests/test_decode_bash_output_672.py
@@ -1,0 +1,41 @@
+"""Unit tests for decode_bash_output() (#672).
+
+decode_bash_output lives in tests/_bash_runner.py next to resolve_bash(),
+the helper that locates a real POSIX bash. These tests are pure-bytes and
+run on every platform -- no bash required -- so the UTF-16LE decode path
+is exercised deterministically even on machines (and CI legs) where the
+exact WSL-launcher corruption does not reproduce.
+"""
+
+from __future__ import annotations
+
+from ._bash_runner import decode_bash_output
+
+
+def test_utf16le_interleaved_blob_decodes_to_readable_text():
+    """Positive control: a UTF-16LE-framed blob -- the shape the WSL
+    launcher or a WindowsApps alias stub writes to a redirected pipe --
+    must come back as readable text, not NUL-interleaved garbage."""
+    raw = "HOOK_DIR=/some/plugin/root/scripts\n".encode("utf-16-le")
+    assert b"\x00" in raw  # the corruption signature is actually present
+    assert decode_bash_output(raw) == "HOOK_DIR=/some/plugin/root/scripts\n"
+
+
+def test_plain_utf8_output_is_returned_unchanged():
+    """Negative control: ordinary UTF-8 output (what Git Bash writes) must
+    pass through untouched -- the decoder must not mangle what is fine."""
+    raw = "HOOK_DIR=/some/plugin/root/scripts\n".encode("utf-8")
+    assert b"\x00" not in raw
+    assert decode_bash_output(raw) == "HOOK_DIR=/some/plugin/root/scripts\n"
+
+
+def test_empty_stream_returns_empty_string():
+    assert decode_bash_output(b"") == ""
+
+
+def test_undecodable_bytes_fall_back_without_raising():
+    """A blob that is neither clean UTF-8 nor a usable UTF-16LE frame must
+    still return a str (with replacement characters), never raise."""
+    raw = b"\xff\xfe\x00\x00garbage\x80\x81"
+    out = decode_bash_output(raw)
+    assert isinstance(out, str)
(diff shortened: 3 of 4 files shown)


fdaviddpt (changes requested):
Thanks for taking this one with a real Windows host -- the root cause (bare `"bash"` resolving to the WSL launcher on `windows-latest`, whose stderr is UTF-16LE) is the right diagnosis, and switching the probe to `resolve_bash()`'s Git Bash is the actual fix. CI runs are approved now.

One correctness bug in `decode_bash_output()` that needs fixing before merge:

The UTF-16LE gate `raw.count(b"\x00") >= len(raw) // 4` is true for **every** input shorter than 4 bytes, because `len(raw) // 4 == 0` and `count >= 0` always holds. A short, perfectly valid UTF-8 output with no NUL in it then gets decoded as UTF-16LE and silently mangled:

```python
>>> decode_bash_output(b"1\n")
'਱'
>>> decode_bash_output(b"ok")
'歯'
>>> decode_bash_output(b"0\n")
'ਰ'
```

`b"1\n"` / `b"0\n"` is exactly the shape a probe that prints an exit status or a counter produces, so this will bite the next caller. Fix: require at least one NUL before attempting the UTF-16LE path, e.g.

```python
nuls = raw.count(b"\x00")
if nuls and nuls >= len(raw) // 4:
```

and please add the negative control to `tests/test_decode_bash_output_672.py` -- a 2-byte UTF-8 input (`b"1\n"`) that must come back unchanged. Per this repo's CLAUDE.md it should be red before the gate change and green after.

Two smaller notes, no change required:
- `test_undecodable_bytes_fall_back_without_raising` only asserts `isinstance(out, str)`; asserting that `"garbage"` survives in the output would make it a real check rather than a no-raise check.
- The changelog fragment and the removed skipif both read well; the `#672` comment replacing the skipif is the right amount of history.

fdaviddpt (approved):
Re-reviewed after 2cfafb2: the NUL gate, the b'1\n' negative control and the 'garbage' assertion are all in. CI 13/13 on the head commit, and the windows-latest log shows the 669 module at 7 passed / 0 skipped, so the previously skipped probe test now runs for real. Approving.

PR #693, opened 2026-09-14
fix(haiku): strip ANTHROPIC_API_KEY from the nested claude -p child env

### LINKED ISSUE #703: An ambient ANTHROPIC_API_KEY out-ranks the operator's login for the nested summarizer, and an exhausted one fails every background save silently
Reported by @Mega-Therion in #693, with the trace that identified it. Filing the symptom so the fix has an issue to close, per that PR's review.

## What happens

`_child_env()` in `pipeline/haiku.py` passes the host environment through to the nested `claude -p` the background summarizer spawns, stripping only the parent-session markers. `ANTHROPIC_API_KEY` goes through with it.

The Claude CLI resolves credentials in a fixed order, and that key out-ranks a `claude.ai` login. So on a host where the operator has both — a login they chose, and an API key set in their shell for some unrelated tool — every nested summarizer call authenticates as the key, not as the login.

When that key's balance is exhausted or rate-limited, every background save fails:

```
[haiku] call exited 1; ...
[haiku] ERROR: call-haiku error: claude exited 1: Credit balance is too low | ⚠ claude.ai connectors are disabled because ANTHROPIC_API_KEY or another auth source is set and takes precedence over your claude.ai login · Unset it to load your organization's connectors
```

Observed 12 times across several days on one installation.

## Why it is hard to see

The operator's own interactive sessions keep working — they run off the login, which is not out-ranked in that context. Only the nested child is affected. In `hook-errors.log` the whole thing reads as `save-session.sh --force exited 1`, with the real cause going to the daily narrative log and nowhere else (#694). Nothing at any layer names the variable.

## Why the obvious fix is not the whole fix

Stripping `ANTHROPIC_API_KEY` unconditionally moves the outage rather than removing it: for an operator whose *only* credential is that key — a normal, documented way to authenticate the CLI, with no `claude.ai` login and no `setup-token` — a strip leaves the child unauthenticated and every background save fails, silently, in exactly the same shape.

So the fix has to distinguish the two populations: strip the ambient key when another credential is actually available, keep it when it is the only one there is. What "available" can mean is itself limited — a `claude.ai` login stored in the macOS Keychain is not visible to this process the way `~/.claude/.credentials.json` is — so whatever automatic rule ships needs an explicit operator override alongside it, and the failure needs to name the variable when it does bite.

## Acceptance

- An ambient `ANTHROPIC_API_KEY` no longer reaches the nested CLI when another credential is available.
- It still reaches it when it is the only credential present.
- An operator can force either behaviour when the automatic rule cannot see their login.
- A failure that this variable plausibly caused says so, in the place the operator is already looking.

[AI-generated]

Mega-Therion:
## Summary
- `ANTHROPIC_API_KEY`, when set anywhere in the host environment, takes precedence over the operator's `claude.ai` login for the nested `claude -p` call the background summarizer spawns — the CLI resolves credentials in that order.
- When that key's balance is exhausted or rate-limited, the background summarizer fails on every session-end with `Credit balance is too low`, silently — the operator's own interactive sessions keep working fine off the login, so nothing about the failure points at this var without reading the child's own stderr.
- Unlike `CLAUDE_CODE_OAUTH_TOKEN`, this was never the child's own credential — it's an ambient host var the operator did not choose for this subprocess. `_child_env()` now strips it, for the same reason it already strips `CLAUDE_PROJECT_DIR`.

## Reproduced in production
Traced via the daily narrative log (`~/.remember/logs/memory-*.log`) to exactly this cause, on a real installation with `ANTHROPIC_API_KEY` set in the environment (for unrelated reasons) and an exhausted balance:
```
[haiku] call exited 1; ...
[haiku] ERROR: call-haiku error: claude exited 1: Credit balance is too low | ⚠ claude.ai connectors are disabled because ANTHROPIC_API_KEY or another auth source is set and takes precedence over your claude.ai login · Unset it to load your organization's connectors
```
12 occurrences across several days on that installation, each showing as an opaque `save-session.sh --force exited 1` warning in `hook-errors.log` with no indication of the actual cause unless you dig into the daily log.

## Test plan
- [x] Added `test_call_haiku_strips_anthropic_api_key`, matching the existing `test_call_haiku_strips_parent_session_env` pattern
- [x] `pytest tests/test_haiku.py` — 80/80 passing (79 pre-existing + 1 new)
- [x] Confirmed manually: `_child_env()` no longer includes `ANTHROPIC_API_KEY` when set in the parent environment
- [x] Diff is minimal (37 insertions, 0 deletions) — no unrelated reformatting

Closes #703.

---
*Maintainer edit (@fdaviddpt): only this closing reference and the note below were added — the description above is the author's own, unchanged. Two commits were pushed to this branch (see the comments below): the conditional-strip rework the review asked for, and a follow-up that stops the config warning echoing a value an operator might have pasted a real key into. #704 was a temporary duplicate of the same work, opened while pushing here was blocked, and is closed in favour of this PR.*

[AI-generated]


fdaviddpt (changes requested):
Thank you for this -- the diagnosis is correct and the evidence is exactly what a report like this needs: the CLI's own stderr, the frequency, and the fact that interactive sessions kept working off the login while the background summarizer did not. That last detail is what makes this failure so hard to see, and you named it precisely.

I'm requesting changes on one thing, plus some repo mechanics.

## The strip is unconditional, and that breaks a different set of installations

`_child_env()` keeps exactly one credential:

```python
_CHILD_ENV_KEEP = frozenset({"CLAUDE_CODE_OAUTH_TOKEN"})
```

For an operator whose *only* credential is `ANTHROPIC_API_KEY` -- no `claude.ai` login, no `setup-token`, which is a normal and documented way to authenticate the Claude CLI -- this patch leaves the nested `claude -p` with nothing to authenticate with. Every background save then fails, silently, surfacing as the same opaque `save-session.sh --force exited 1` you spent days tracing. That is the failure class this PR exists to remove, relocated onto a different population rather than removed.

Worth noting that `ANTHROPIC_API_KEY` appears nowhere in `docs/`, `README.md` or `scripts/` today, so neither behaviour is documented and we have no data on how many installs depend on it. That argues for caution rather than for assuming nobody does.

What I'd like instead: **strip it only when another credential is actually available.** Concretely, keep `ANTHROPIC_API_KEY` when none of `CLAUDE_CODE_OAUTH_TOKEN` (in the child env), `REMEMBER_OAUTH_TOKEN` or `haiku.oauth_token` is present; strip it when one of them is, since that is the case your report describes -- an ambient key out-ranking a login the operator actually chose. `_inject_configured_oauth_token()` already resolves the configured-token half, so the check has somewhere natural to live.

## Test coverage

The new test is a good positive control. It needs its negative twin, and this repo asks for that pairing explicitly (CLAUDE.md: "a negative assertion needs a positive control", and the same in reverse): a case where `ANTHROPIC_API_KEY` is the sole credential and therefore **survives**. Without it, a future refactor that strips unconditionally again passes the suite.

`test_call_haiku_keeps_oauth_token` is the pattern to mirror.

## Repo mechanics

- **A changelog fragment is a merge gate here.** Add `changelog.d/<issue>.fixed.md` -- one line, user-facing. The `oss changelog` check fails without it.
- **No linked issue.** A `Closes #N` line binds the PR to an issue; a bare `#N` does not. Happy for you to file one describing the symptom, or I can.
- CI on this PR was waiting on first-time-contributor approval; I've approved it, so the twelve legs will report shortly.

## Separately, and not your problem to fix

Your logs showed the real cause in `memory-*.log` while `hook-errors.log` carried only the exit code. That propagation gap is a bug of ours independent of this change, and it is why this took days instead of minutes to find -- filed as #694 with your trace as the evidence. Thanks for surfacing it.

fdaviddpt:
I've finished this one, in #704. Your commit is the first of the two there, unchanged and under your name — #704 adds the condition my review asked for rather than replacing your work, and the issue it closes (#703) is filed with your trace as the evidence.

Two things worth saying plainly.

**Why I did not push to your branch.** That was my first move — `maintainer_can_modify` is set on this PR, so it was available — and my tooling declined to write to someone else's fork. Rather than ask you to wait on that, I put the same two commits on a branch here. If you would rather carry it yourself, say so and I will close #704 and hand you the diff; the work is genuinely yours to finish if you want it.

**What changed from your patch.** The strip is now conditional. `ANTHROPIC_API_KEY` is dropped when another credential is visible — `CLAUDE_CODE_OAUTH_TOKEN`, a configured OAuth token, or a login at `~/.claude/.credentials.json` — and kept when it is the only credential on the host, because for an operator who authenticates the CLI with that key alone a strip leaves the child with nothing and every background save fails in exactly the shape you spent days tracing. A `claude.ai` login in the macOS Keychain is invisible to this process, so there is a `haiku.anthropic_api_key` config key (`auto`/`keep`/`strip`) for the operators the automatic rule cannot see. And a credential-shaped failure now names the variable in the error itself, which is the part of your report I kept coming back to: the mechanism was findable, but only by someone willing to read the child's stderr in a daily log.

The negative twin — the key surviving when it is the only credential — is the test that fails against the original patch, and it is there so a future refactor cannot quietly strip again.

Closing this in favour of #704. Thank you for the report; the trace in it is what made the whole thing tractable, and the framing in your description (interactive sessions fine, background summarizer dead) is what pointed at the credential order in the first place.


[AI-generated]

fdaviddpt:
Correction to my comment above: the push to this branch went through on a retry, so **the work is here, not in #704** — I've closed that one as a duplicate. Your branch now carries three commits: yours, the conditional-strip rework, and one more described below. Nothing of yours was rewritten; both of mine sit on top of `040fa8b`.

The third commit is worth naming, because CodeQL caught it on the second push and it was right to. My warning for an unrecognised `haiku.anthropic_api_key` value quoted the value back — and that key's *name* invites someone to paste an actual API key into the field. A typo would then have written a live credential to the daily log in clear text. It now reports the length and type, the legal values and the fallback, and says the value is withheld and why. The test asserts the absence, with a positive control (a recognised value warns about nothing) so "warns on a bad value" cannot quietly become "warns on everything".

I also added `Closes #703` to the description and a short note marking that edit — your text above is untouched.

CI is running on the branch now. Once it is green this is ready to merge as yours.


[AI-generated]

fdaviddpt:
Merging this — the change my earlier review asked for is in the branch, so treat that CHANGES_REQUESTED as answered rather than standing. (I could not clear the review itself: approving a pull request is an action my tooling holds back from me, which seems right. The maintainer's decision to merge is on the record here instead.)

State at the merge, named rather than summarised:

- **15 legs passed, 0 failed** on `c90b49b`.
- **One CodeQL check run is `neutral`**, not green: "1 configuration not found: `/language:actions`". That is code scanning declining to compute the PR-vs-main alert diff for the actions configuration — not a failed analysis. Both analyses on this sha (`Analyze (python)`, `Analyze (actions)`) concluded success, and this branch touches no workflow files, so there is no actions-language alert it could have introduced.
- **The base is 7 commits behind `main`.** The green tally is therefore evidence about this branch's merge-base, not about the tree that lands. Merging on that basis deliberately; `main`'s own run is what settles it, and anything that surfaces there is a follow-up rather than a reason to hold this.
- The first CI attempt on this branch produced two runs with **zero jobs and a `failure` conclusion** — nothing ran. A re-run produced the full matrix and passed. Transient, GitHub-side.

Thank you again for the report, the trace, and the patch. Landing as yours.


[AI-generated]

fdaviddpt (approved):

Issue #695, opened 2026-09-15
safe_eval drops every bridge variable containing 'I' under tr_TR locale — no save ever completes


erkdgn:
## Summary

Under a Turkish locale (`LC_COLLATE=tr_TR.UTF-8`), **no save ever completes**. `safe_eval` in `scripts/log.sh` matches `KEY=VALUE` lines with the bracket range `^([A-Z_][A-Z0-9_]*)=`. Bash evaluates that range with the locale's collation, and tr_TR collation does not place `I` inside `[A-Z]`. So every bridge variable whose name contains an `I` is silently skipped:

| Variable | `LC_ALL=tr_TR.UTF-8` | `LC_ALL=C.UTF-8` |
|---|---|---|
| `EXTRACT_FILE`, `POSITION`, `SKIP_LINES`, `UNREAD_SIDECAR_UNREADABLE` | **not assigned** | assigned |
| `HUMAN_COUNT`, `EXCHANGE_COUNT`, `ENVELOPE`, `ENVELOPE_HAS_UNMAPPED_STEP` | assigned | assigned |

The counts still arrive, so `save-session.sh` passes the "0 exchanges" gate and reaches `build-prompt` with an empty `EXTRACT_FILE`:

```
[extract] 270 exchanges (36 human)
[error] FAILED at line 678 (exit 1)
```

```
File ".../pipeline/shell.py", line 178, in cmd_build_prompt
    with open(extract_file, encoding="utf-8", errors="replace") as f:
FileNotFoundError: [Errno 2] No such file or directory: ''
```

On one project `hook-errors.log` holds 5,071 of these, plus 799 `ValueError: invalid literal for int() with base 10: ''` (empty `POSITION`/`SKIP_LINES`). `/remember:doctor` reports `FAIL No save has ever completed for this project`.

## Reproduction

```bash
LC_ALL=tr_TR.UTF-8 bash -c 'for k in POSITION EXTRACT_FILE HUMAN_COUNT; do
  l="$k=x"; [[ "$l" =~ ^([A-Z_][A-Z0-9_]*)=(.*)$ ]] && echo "match $k" || echo "NO MATCH $k"; done'
# NO MATCH POSITION
# NO MATCH EXTRACT_FILE
# match HUMAN_COUNT
```

[...]

PR #698, opened 2026-09-15
diag(#695): which bracket-range shapes diverge under Turkish collation

fdaviddpt:
**Throwaway diagnostic. Do not merge.** Red on purpose -- the table is the
output, same shape as the #689 diagnostics.

#695 established one divergence: `[A-Z]` does not match `I` under glibc
`tr_TR`. Every other bracket range in `scripts/` is spelled differently --
`[a-zA-Z]`, `[A-Za-z0-9._-]`, the negated `case` form, the
`${v//[!a-zA-Z0-9]/-}` substitution form -- and whether those also drop `i`/`I`
decides how many sites actually need `LC_ALL=C`.

macOS cannot answer it: all six shapes behave identically there across `C`,
`en_US.UTF-8` and `tr_TR.UTF-8`, because darwin's libc does not collate ranges
the way glibc does. So this gets the matrix off a Linux runner once, and the
branch is deleted.

Part of #695

[AI-generated]


fdaviddpt:
Matrix read and recorded: only `[[ =~ ]]` collates on glibc; `case` patterns and `${v//[!...]/}` compare bytes and did not move under tr_TR. Both failure directions are real in the regex path -- `[A-Z]` loses `I`, `[A-Za-z]` gains `é`. Carried into #699 (PR body, and the docstrings of tests/test_locale_ranges_695.py and tests/test_id_guards_locale_695.py), which is scoped to what this measured rather than to what I assumed before running it. Throwaway branch, deleting.

Issue #694, opened 2026-09-14
A failed background save reports only `exited 1` in hook-errors.log; the real cause goes to the daily log and nowhere else


fdaviddpt:
## Logging gap: the real cause of a failed background save is written only to the daily narrative log, never to hook-errors.log

Surfaced by @Mega-Therion while diagnosing #693. Their installation failed every background summarize for several days, and the only thing `hook-errors.log` had to say about it was:

```
save-session.sh --force exited 1
```

The actual cause was sitting in `~/.remember/logs/memory-*.log` the whole time:

```
[haiku] ERROR: call-haiku error: claude exited 1: Credit balance is too low | ...ANTHROPIC_API_KEY or another auth source is set and takes precedence over your claude.ai login
```

`_failure_detail()` (`pipeline/haiku.py:404`) already does the hard part -- it digs the real message out of the CLI's stdout JSON, which is where auth and billing failures live (#129/#152). That message reaches the daily log. It does not reach the log an operator actually looks at when saves stop working, which surfaces only a return code.

## Why this is its own issue

A return code with no cause is the shape this project exists to avoid: something ran, produced a clean-looking failure line, and said nothing about why. 12 occurrences across several days were needed before the reporter went digging in a second log file to find a message that had been written every single time.

This is independent of whichever fix #693 lands on. Any nested-call failure -- expired OAuth, rate limit, exhausted balance, a CLI that will not start -- has the same shape: diagnosed correctly, recorded in one place, and reported as an exit code in the other.

## Shape of a fix

`save-session.sh`'s failure path should carry the child's own failure detail into `hook-errors.log`, not just its exit status. The detail is already computed and already length-capped (`_FAILURE_DETAIL_MAX`, 500 chars), so this is about propagation, not about producing new information.

A test wants both directions: a failing nested call whose reason reaches `hook-errors.log` (must-fire), and an ordinary successful save that writes nothing there (must-not-fire).

[AI-generated]

PR #704, opened 2026-09-16
fix(#703): strip the ambient ANTHROPIC_API_KEY on evidence, keep it when it is the only credential

### LINKED ISSUE #703: An ambient ANTHROPIC_API_KEY out-ranks the operator's login for the nested summarizer, and an exhausted one fails every background save silently

fdaviddpt:
Closes #703. **Supersedes #693 — the diagnosis and the original patch are @Mega-Therion's**, and their commit is the first of the two here, unchanged. What this adds is the condition the review asked for.

## Why the strip could not stay unconditional

`ANTHROPIC_API_KEY` out-ranks a `claude.ai` login in the CLI's credential order, so an ambient key becomes the nested summarizer's credential and an exhausted one fails every background save while the operator's interactive sessions carry on working. That is #703, and dropping the key fixes it.

But authenticating the CLI with `ANTHROPIC_API_KEY` alone — no login, no `setup-token` — is a normal, documented setup, and for those installs a strip leaves the child with no credential at all: every save failing, silently, in the same shape that took the reporter days to trace. The fix and the bug are the same outage pointed at different people.

So the key is dropped on evidence:

| what is visible | ambient key |
|---|---|
| `CLAUDE_CODE_OAUTH_TOKEN` from the host | stripped |
| a configured `haiku.oauth_token` / `REMEMBER_OAUTH_TOKEN` | stripped |
| a login at `~/.claude/.credentials.json` (honours `CLAUDE_CONFIG_DIR`) | stripped |
| nothing else | **kept** — it is the only credential there is |

## The limit, stated rather than papered over

A `claude.ai` login held in the **macOS Keychain is invisible** to this process, and probing the keychain would prompt the operator and read a secret we have no business reading. So `auto` keeps the key for those operators — today's behaviour, not a new failure — and they say so once:

```json
{ "haiku": { "anthropic_api_key": "strip" } }
```

`auto` (default), `keep`, `strip`. An unrecognised value falls back to `auto` **and says so in the daily log**, rather than grading silently as one of the two behaviours.

## Discoverability

The other half of #703 is that nothing named the variable. A failure that looks like a credential failure now carries it into the `RuntimeError`, which `save-session.sh` surfaces to `hook-errors.log` — where the operator is already looking, instead of the daily log alone (#694):

```
claude exited 1: Credit balance is too low -- note that ANTHROPIC_API_KEY was set in this
environment and was passed to the nested CLI, where it OUT-RANKS a claude.ai login; if that
key is exhausted or wrong, this is what failed. Set `haiku.anthropic_api_key` to "strip" in
config.json to keep it out of the summarizer, or "keep" to silence this note (#703)
```

The marker list is deliberately narrow. A hint that fires on every failure points unrelated outages at an innocent variable, which is the misdirection this issue is about.

## Tests

Every strip case has its keep twin, because this is a repo rule and here it is load-bearing: an unconditional strip and an unconditional keep would each pass half the suite.

- strips when the host supplied `CLAUDE_CODE_OAUTH_TOKEN` / when an OAuth token is configured / when a login file exists
- **keeps when it is the only credential** — the twin that fails against @Mega-Therion's original patch, and against any future refactor that strips again
- policy `strip` overrides an invisible login; policy `keep` overrides the strip; an unrecognised value warns and falls back
- the failure hint fires on a credit error with the key set, and does **not** fire with no key set (negative control) or on a prompt-length failure with the key set (wrong-cause control)

The fixture relocates `HOME`, `REMEMBER_DIR` and the token env vars: without it the decision under test is the developer's own login, and the result flips between a laptop and CI.

`pytest`: **2496 passed, 58 skipped** locally (macOS). The Windows and Linux legs are reasoned, not observed, from here — nothing in this change is platform-specific beyond `expanduser`, which the fixture pins on both shapes via `HOME` and `USERPROFILE`.

Docs: `docs/configuration.md` gets the row, `config.example.json` ships `"anthropic_api_key": "auto"`, and `tests/test_config_contract.py` records that this key is read in Python rather than through `config()`.

[AI-generated]


fdaviddpt:
Closing as a duplicate. This existed only because pushing to the author's own branch was blocked at the time; the retry succeeded, so the identical commits now sit on #693, where they belong — @Mega-Therion reported the bug, diagnosed it and wrote the first commit, and their PR should be the one that lands.

Same two commits, same tests, no work lost. Continue at #693.


[AI-generated]

Issue #690, opened 2026-09-14
session-start-hook.sh cannot be profiled with `bash -x`: bootstrap-dirs.sh:364 redirects fd 2 mid-run and the trace stops silently


fdaviddpt:
## What happens

`scripts/bootstrap-dirs.sh:364` runs:

```sh
exec 2>> "$REMEMBER_DIR/logs/hook-errors.log"
```

From that line on, **everything written to fd 2 goes to the hook's own error log** -- including an `xtrace` stream, if the profiler put one there. `bash -x scripts/session-start-hook.sh` therefore produces a trace that covers only the part of the run before `bootstrap-dirs.sh` is sourced, and nothing anywhere says the rest was lost.

Measured while building the #660 diagnostic, on macOS, one run:

```
wall 0.71s | traced span 0.07s | 238 traced lines
```

90% of the run was missing. With `BASH_XTRACEFD` pointed at a dedicated fd, the same run gives `wall 0.71s | traced span 0.66s | 1372 traced lines`.

## Why it matters

A truncated trace does not look truncated. It looks like a complete profile of a fast hook, and every number derived from it -- fork counts, per-step attributions, "the largest gap is X" -- describes a fraction of the run while reading as though it describes all of it. This is the repo's own cardinal failure shape (a thing that ran, produced a clean result, and silently covered a fraction of what it claimed) sitting inside a diagnostic aid.

It is also not hypothetical: #660 has been reasoned about for days using `PS4='+$EPOCHREALTIME ' bash -x` measurements from a reporter, and unless their trace was on a non-default fd it stopped at this same line.

## Not a proposal to remove the redirect

The redirect exists for a reason -- a hook's stderr must not leak into the host's transcript (#643 and the stderr-leak lane). The ask is that it stop being invisible:

- leave the redirect alone when `BASH_XTRACEFD` is already set to something other than 2, or when an explicit opt-out (`REMEMBER_TRACE=1`) is present; and/or
- document it at the redirect and in `docs/diagnostics.md`, so the next person to profile this hook is told where their trace went rather than discovering it by arithmetic.

A positive control belongs in whichever shape lands: a traced run whose span is within some margin of its wall clock, so a future redirect that swallows the trace again fails a test rather than a person.

[AI-generated]

fdaviddpt:
Two platform floors the fix in PR #708 does not yet clear. Recording them here rather than
only on the pull request, so they survive if that branch stalls. Both are read out of run
[...]

PR #733, opened 2026-09-19
fix(#712): skip session-start-hook.sh's stdout buffer under an active trace

### LINKED ISSUE #712: bash -x session-start-hook.sh produces zero stdout bytes on windows-latest (round 2 of #690)
Split out of #690 / PR #708 so it survives independently of that branch and is visible on the tracker rather than only inside an `xfail` marker.

## What happens

On the `windows-latest` CI runner, `test_the_real_hook_traces_past_the_bootstrap` in `tests/test_trace_not_swallowed_690.py` fails: `bash -x scripts/session-start-hook.sh` exits 0 but produces zero stdout bytes. The six other tests in the same file pass on the same runner, so this is not a `bash` resolution or `PATH` handling problem -- something specific to this end-to-end path (or to how the trace/output is captured) does not work on that platform.

This is the exact test that pins #690's own promise -- a `bash -x` trace of the real hook reaching past `bootstrap-dirs.sh`'s fd-2 redirect -- so on Windows that promise is currently unverified, not merely untested.

## What is not yet known

Whether the `_REMEMBER_CTX_FILE` buffer redirect (`scripts/session-start-hook.sh:1527-1542`, `1969-1972`) never engages under trace on Git Bash, or engages and the flush branch is never reached. This needs a live Windows runner (or an equivalent Git Bash environment) to step through.

## Current state

PR #708 (which fixed the macOS bash-3.2 regression in the same file, PR unaffected by this issue) marks the Windows case `xfail(sys.platform == "win32", strict=False, raises=AssertionError)` rather than a silent skip, so it stays visible until fixed (flips to XPASS, which is the signal to remove the marker) rather than quietly passing. #690 itself is being kept open rather than closed by that PR, since this is the same end-to-end guarantee the issue was filed for, just on a different platform.

## Acceptance

The xfail marker on `test_the_real_hook_traces_past_the_bootstrap` (currently scoped to `sys.platform == "win32"`) can be removed because the test passes for real on `windows-latest`, and #690 can then be closed once this is fixed.

[AI-generated]

fdaviddpt:
Round-2 windows-latest leg of #690. `bash -x scripts/session-start-hook.sh` was observed exiting 0 with zero stdout bytes on windows-latest in `test_the_real_hook_traces_past_the_bootstrap`, while six neighbouring bootstrap-dirs.sh-only tests in the same file pass on the same runner. No Windows runner was available to this lane (or to the original issue filer) to step through the actual Git-Bash fd mechanism directly, so the fix below is REASONED, not OBSERVED, on the one platform it targets.

`session-start-hook.sh` buffers its own stdout into a temp file (`exec 3>&1; exec > "$_REMEMBER_CTX_FILE"`) so a plugin-promo `systemMessage` JSON can wrap the output, then restores the real fd and `cat`s the buffer back at the end. This adds a guard that skips the whole buffer/fd-swap whenever an active xtrace (`case "$-" in (*x*))`) or `REMEMBER_TRACE=1` is set -- mirroring the existing, already-shipped precedent in `bootstrap-dirs.sh`, which already stands its own fd 2 redirect down under the identical two conditions (don't shadow output an operator deliberately went looking for). When the guard fires, `_REMEMBER_CTX_OK` stays empty and the hook falls back to the exact code path already proven safe by `test_unwritable_ctx_buffer_does_not_burn_the_cooldown` (buffer unwritable -> print live, no promo, no cooldown marker burned). The only visible cost of the guard firing is that the plugin-promo banner does not show under an active trace; the memory context itself is never affected.

Also updates the stale `xfail` reason text on `test_the_real_hook_traces_past_the_bootstrap` to describe the fix and point at the new test class, while leaving the xfail itself in place (`strict=False`, scoped to `sys.platform == "win32"`, `raises=AssertionError`) since there is no Windows runner available in this environment to confirm it flips to XPASS -- that confirmation has to come from the next windows-latest CI run.

## New tests

`tests/test_trace_not_swallowed_690.py::TestCtxBufferSkippedUnderTrace`, using the plugin-promo `systemMessage` JSON as a cross-platform-observable proxy for whether the buffer engaged:

- `test_positive_control_promo_shows_without_trace` -- must-fire control: untraced, buffer writable, promo-eligible fixture -> systemMessage present.
- `test_buffer_skipped_under_bash_x` -- the actual guard: `bash -x` active -> systemMessage absent.
- `test_buffer_skipped_under_remember_trace` -- `REMEMBER_TRACE=1` -> systemMessage absent, the same explicit opt-out `bootstrap-dirs.sh` already honours.

## Testing

RED (`scripts/session-start-hook.sh` reverted, new tests kept): `pytest tests/test_trace_not_swallowed_690.py -k TestCtxBufferSkippedUnderTrace -q --no-cov` -> 2 failed, 1 passed (positive control passed; both guard tests failed against the old code).

GREEN: `pytest tests/test_trace_not_swallowed_690.py -q --no-cov` -> 10 passed, 1 skipped (pre-existing, unrelated `BASH_XTRACEFD`-version skip). `pytest tests/test_plugin_promo_574.py tests/test_session_start_fork_tax_665.py tests/test_session_start_spawn_reduction_679.py -q --no-cov` -> 27 passed, no regressions in neighbouring `session-start-hook.sh` suites. Full `pytest` (the whole `test_command`) was not run locally, narrowed per policy to touched files plus the guard-derived set.

Closes #712

[AI-generated]


fdaviddpt:
Follow-up commit `22833e7` pushed after CI went red on windows-latest (job 105918977745).

The failing leg was this PR's own new positive control (`test_positive_control_promo_shows_without_trace`), which failed with `assert "systemMessage" in ''` and the job's own stderr showing `session-start-hook.sh: line 213: ./resolve-paths.sh: No such file or directory`.

Root cause: the test invoked the real hook as `bash [..., str(SESSION_START)]`. On Windows, plain `str()` on a `WindowsPath` renders a pure-backslash path with no forward slash at all. `session-start-hook.sh`'s self-location logic (`_HOOK_DIR="${BASH_SOURCE[0]%/*}"`) has no fallback for that shape, so with zero `/` in `BASH_SOURCE[0]` the strip is a no-op and `_HOOK_DIR` becomes `.`  -- `source "$_HOOK_DIR/resolve-paths.sh"` then fails from whatever directory pytest happens to be running in, and `resolve-paths.sh`'s own `|| exit 0` swallows that failure silently. This is very likely also the exact original #712 symptom (`bash -x session-start-hook.sh` exiting 0 with zero stdout bytes) -- a test/harness path-normalization bug, not a production bug (`hooks.json` always invokes via a literal path with at least one forward slash regardless of platform).

This is a known bug class in this repo: `tests/test_windows_native_hook_cwd_448.py` documents hitting and fixing the identical thing, and `tests/test_hooks_json.py` establishes `.as_posix()` as the standing convention. All four `bash [..., str(SESSION_START)]` call sites in `tests/test_trace_not_swallowed_690.py` now use `.as_posix()` instead.

The original fd-swap/trace guard from `16598ee` is kept as a harmless, precedent-matching improvement (mirrors `bootstrap-dirs.sh`'s existing convention), but the path-normalization fix is now believed to be the actual root cause of the CI symptom. `changelog.d/712.fixed.md` updated to describe both. Not independently reproduced on a Windows runner -- REASONED, confirmed by the next windows-latest CI run rather than locally.

Note: this branch's own CI may still show unrelated red from #734 (a stale `_SANCTIONED_DIVERGENCE` entry on `main` at 7ba7609, tracked separately) until that lands -- unrelated to this fix.

[AI-generated]

